PR #18, opened 2023-06-08
Adds missing url

StrongestNumber9:


StrongestNumber9:
Self approving as this is trivial update

PR #315, opened 2024-09-10
Split maven/github releases to two different workflows

StrongestNumber9:


StrongestNumber9:
Implemented properly via #470

PR #318, opened 2024-09-11
Pattern acceleration support

elliVM:
## DRAFT requires changes to work with new version of PTH-03 version 8.0.0

PR moved from internal repository 

## Adds pattern parameter to match pth_06 pattern acceleration schema changes. pattern value is fetched from config

- Pattern value is added to inserts into filtertype table
- Pattern is added to the FK to filtertype in created bloom filter tables 
- Refactoring to comply with company standards
- Object equality and testing


src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
```diff
@@ -100,7 +102,7 @@ public TeragrepBloomStep(
         this.outputCol = outputCol;
         this.estimateCol = estimateCol;
 
-        if (mode == BloomMode.ESTIMATE) {
+        if (mode == BloomMode.ESTIMATE || mode == BloomMode.AGGREGATE) {
```
eemhu, line 105: In-constructor code is not ideal, but as far as I remember there is an issue in StepList where properties cannot be added in any other place. Perhaps a new issue?

src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
```diff
@@ -76,16 +77,17 @@ public enum BloomMode {
...
-    private String estimateCol;
+    private final String inputCol;
+    private final String outputCol;
+    private final String estimateCol;
 
     // Bloom filter consts
     public final static String BLOOMDB_USERNAME_CONFIG_ITEM = "dpl.pth_10.bloom.db.username";
     public final static String BLOOMDB_PASSWORD_CONFIG_ITEM = "dpl.pth_10.bloom.db.password";
     public final static String BLOOMDB_URL_CONFIG_ITEM = "dpl.pth_06.bloom.db.url";
     public final static String BLOOM_NUMBER_OF_FIELDS_CONFIG_ITEM = "dpl.pth_06.bloom.db.fields";
```
eemhu, line 88: could these be private?

elliVM:
Added support for dpf-03 regex tokenizer
Bloom filter table tokens are filtered using regex match against a pattern provided by user

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/FilterTypes.java
```diff
@@ -110,6 +108,7 @@ public Map<Long, Long> bitSizeMap() {
 
     public String pattern() {
         final String pattern;
+        String BLOOM_PATTERN_CONFIG_ITEM = "dpl.pth_06.bloom.pattern";
```
eemhu, line 111: these could be final, or better yet, object of their own

src/main/java/com/teragrep/pth10/steps/tokenizer/RegexTokenizerUDF.java
```diff
@@ -0,0 +1,95 @@
...
+        final UserDefinedFunction regexUDF = functions
+                .udf(
+                        new com.teragrep.functions.dpf_03.RegexTokenizerUDF(),
+                        DataTypes.createArrayType(DataTypes.BinaryType, false)
+                );
+
+        if (this.format == AbstractTokenizerStep.TokenizerFormat.BYTES) {
+            return dataset
+                    .withColumn("regex_column", functions.lit(pattern))
+                    .withColumn(outputCol, regexUDF.apply(functions.col(inputCol), functions.col("regex_column")))
```
eemhu, line 86: I think it should be possible to use `functions.lit(pattern)` as the second parameter of regexUDF, could avoid creating and dropping `regex_column`

src/test/java/com/teragrep/pth10/TokenizerTest.java
```diff
@@ -142,4 +156,52 @@ public void tokenize3() {
...
+    )
+    public void testRegexTokenizerNotSupportedFormat() {
+        Properties properties = new Properties();
+        properties.put("dpl.pth_06.bloom.pattern", "testRegex");
+        Config config = ConfigFactory.parseProperties(properties);
+
+        TokenizerStep step = new TokenizerStep(config, AbstractTokenizerStep.TokenizerFormat.STRING, "_raw", "result");
+        SparkSession spark = streamingTestUtil.getCtx().getSparkSession();
+        Assertions
+                .assertThrows(UnsupportedOperationException.class, () -> step.get(spark.read().schema(testSchema).csv(spark.emptyDataset(Encoders.STRING()))));
```
eemhu, line 173: Might be useful to assert the exception message as well

src/test/java/com/teragrep/pth10/TokenizerTest.java
```diff
@@ -142,4 +156,52 @@ public void tokenize3() {
...
+        );
+        SparkSession spark = streamingTestUtil.getCtx().getSparkSession();
+        Dataset<Row> result = step.get(spark.read().schema(testSchema).json(regexTestFile));
+        // first row value: 127.0.0.0
+        List<Object> list = result.select("bytetokens").first().getList(0);
+        List<String> tokens = new ArrayList<>();
+        for (Object o : list) {
+            tokens.add(new String((byte[]) o, StandardCharsets.UTF_8));
+        }
+        Assertions.assertEquals(4, list.size());
```
eemhu, line 202: While the expected result's size=4, with values 127, 0, 0 and 0, maybe remove duplicates from list to make sure there is only 127 and 0, because here the other two values could be anything else

kortemik:
please rebase

src/test/java/com/teragrep/pth10/TokenizerTest.java
elliVM: switched list to set with expect result size 2 and values 127, 0

src/test/java/com/teragrep/pth10/TokenizerTest.java
elliVM: added exception message assertion

src/main/java/com/teragrep/pth10/steps/tokenizer/RegexTokenizerUDF.java
elliVM: correct, implemented literal column straight to function

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/LazyConnection.java
```diff
@@ -86,6 +82,7 @@ public synchronized Connection get() {
     private String connectionUsername() {
 
         String username;
+        String BLOOMDB_USERNAME_CONFIG_ITEM = "dpl.pth_10.bloom.db.username";
```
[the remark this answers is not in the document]
elliVM: made final

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/FilterTypes.java
elliVM: made final. Object of some sorts would indeed be better but chose not to implement in this PR scope

elliVM:
switched to another PR

PR #336, opened 2024-09-17
Reduce logging in tests by using slf4j-api instead of slf4j-simple, i…

### LINKED ISSUE #305: Reduce spark logging in tests
**Description**

Reduce spark logging in tests

**Use case or motivation behind the feature request**

It prints a lot of irrelevant logs which slows down usage and drowns possibly interesting messages.
I think the default logging level can be set to WARN+ for spark components long as it can be changed on demand

**Related issues**

**Additional context**

Any build log from actions

StrongestNumber9:
…ntroduce new log4j2-test.xml

Note: Currently prints warning about multiple slf4j providers, but only once.

```
[INFO] Running com.teragrep.pth10.ReplaceTransformationTest
SLF4J: Class path contains multiple SLF4J providers.
SLF4J: Found provider [org.apache.logging.slf4j.SLF4JServiceProvider@2dfaea86]
SLF4J: Found provider [org.slf4j.jul.JULServiceProvider@15888343]
SLF4J: See https://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual provider is of type [org.apache.logging.slf4j.SLF4JServiceProvider@2dfaea86]
```

Fixes #305

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -241,9 +241,8 @@
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-simple</artifactId>
-      <version>1.7.36</version>
-      <scope>test</scope>
+      <artifactId>slf4j-api</artifactId>
+      <version>2.0.16</version>
     </dependency>
   </dependencies>
   <build>
@@ -396,6 +395,8 @@
             <!-- readme -->
             <exclude>README.adoc</exclude>
             <exclude>README.md</exclude>
+            <!-- test logging config -->
+            <exclude>src/test/resources/log4j2-test.xml</exclude>
             <!-- dummy files -->
             <exclude>src/test/resources/hdfslist/**</exclude>
             <exclude>src/test/resources/csv/**</exclude>
--- a/src/test/resources/log4j2-test.xml
+++ b/src/test/resources/log4j2-test.xml
@@ -9,7 +9,13 @@
     <Logger name="com.teragrep" level="info">
       <AppenderRef ref="Console"/>
     </Logger>
-    <Root level="warn">
+    <Logger name="org.apache.spark" level="warn">
+      <AppenderRef ref="Console"/>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -241,9 +241,8 @@
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-simple</artifactId>
-      <version>1.7.36</version>
-      <scope>test</scope>
+      <artifactId>slf4j-api</artifactId>
+      <version>2.0.16</version>
```
kortemik, line 245: has no scope, is it possible to use test scope?

src/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,16 @@
...
+  <Appenders>
+    <Console name="Console" target="SYSTEM_OUT">
+      <PatternLayout pattern="%5p [%d] [%C] ({%t} %F[%M]:%L) - %m%n"/>
+    </Console>
+  </Appenders>
+  <Loggers>
+    <Logger name="com.teragrep" level="info">
+      <AppenderRef ref="Console"/>
+    </Logger>
+    <Root level="warn">
```
kortemik, line 12: is it possible to use default info but org.apache.spark.internal.Logging to warn?

StrongestNumber9: Seems like so, updated

pom.xml
StrongestNumber9: Not possible to use test scope unless org.slf4j Logger gets provided via some route

DedupTransformation.java:[60,17] package org.slf4j does not exist
DedupTransformation.java:[68,26] cannot find symbol
[ERROR]   symbol:   class Logger
[ERROR]   location: class com.teragrep.pth10.ast.commands.transformstatement.DedupTransformation

kortemik (approved):

Issue #320, opened 2024-09-13
Disabled SpathTransformation tests


Abigael-JT:
**Description**

Spath test that tests spath command with json data created from eval  (spathTestEvaledJsonData()) is disabled with the explanation: " Seems like struck unescapes in eval, and the unescaped _raw is given to spath"

Similarly, spathTestJsonNoPath()  test is also disabled  with a comment " output without path is invalid syntax"

This needs to be investigated if these issues have been resolved.

It will be worked on in issue https://github.com/teragrep/pth_10/issues/295

Abigael-JT:
PR created, peer review done, now in architect review pipeline
https://github.com/teragrep/pth_10/pull/348

Issue #295, opened 2024-08-29
enable disabled tests


kortemik:
**Description**
currently there are tests that are disabled for various reasons, please go them through and enable where possible or propose a workaround. please create separate issue for each of the tests and link here.

PR #348, opened 2024-09-25
Enabled disabled spath tests

Abigael-JT:
changes made to the tests:

spathTestJsonNoPath(): Added missing assertions and updated the test query

testSpathEvaledJsonData(): Removed output parameter from the query since its not supported when path is not specified

related issues:
https://github.com/teragrep/pth_10/issues/320
https://github.com/teragrep/pth_10/issues/295


eemhu (approved):
Looks good, ready for architect review

Issue #330, opened 2024-09-16
Eval spath() with JSON/XML tests


Abigael-JT:
**Description**

These tests are disabled because they were broken by spath udf changes.

parseEvalSpathJSONCatalystTest()
parseEvalSpathXMLCatalystTest()

Will be investigate in Issue #295 if this is still the case.

PR #355, opened 2024-10-01
Added backticks to spath keys and enabled the tests

Abigael-JT:
changes made:
- Added backticks to spath keys to escape dots in the path 
- Enabled eval spath test 
- changed expected column arrangement in aggregateAfterHdfsLoadTest

Related ticket
* #330


src/test/java/com/teragrep/pth10/evalTest.java
```diff
@@ -2609,7 +2609,7 @@ public void parseEvalSpathJSONCatalystTest() {
     // Test spath() with XML
     // //person[age=30]/name/text()
     // FIXME broken due to spath udf changes
-    @Disabled
+    //@Disabled
```
elliVM, line 2612: Could this line be removed instead of commenting out?

elliVM (changes requested):
Applying of backticks could be done by an object that wraps the String with methods for formatted and original String values. Alternatively formatted Strings could be created at the start of the method with names like `stringWithBackticks` to clear the code a bit. This solution is also fine those changes are up to you.

### Check this please
Seems like some code format was changed. Please run  `mvn spotless:apply` to check for changes so it doesn't pop up on other branches. Otherwise looks good

Abigael-JT:
> Applying of backticks could be done by an object that wraps the String with methods for formatted and original String values. Alternatively formatted Strings could be created at the start of the method with names like `stringWithBackticks` to clear the code a bit. This solution is also fine those changes are up to you.
> ### Check this please
> 
> Seems like some code format was changed. Please run `mvn spotless:apply` to check for changes so it doesn't pop up on other branches. Otherwise looks good

Removed commented out code and added objects to escape and unescape backticks in spath keys, also added tests for those new objects. We also changed one of the spath test  to only filter out the duplicates and now there is only one assertion.

src/main/java/com/teragrep/pth10/steps/spath/SpathEscapedKey.java
```diff
@@ -0,0 +1,46 @@
+package com.teragrep.pth10.steps.spath;import java.io.Serializable;
```
elliVM, line 1: Add license. `mvn spotless:apply` should create missing licenses

src/main/java/com/teragrep/pth10/steps/spath/SpathUnescapedKey.java
```diff
@@ -0,0 +1,41 @@
+package com.teragrep.pth10.steps.spath;
```
elliVM, line 1: Missing license

src/test/java/com/teragrep/pth10/SpathEscapedKeyTest.java
```diff
@@ -0,0 +1,54 @@
+package com.teragrep.pth10;
```
elliVM, line 1: Missing license

src/test/java/com/teragrep/pth10/SpathUnscapedKeyTest.java
```diff
@@ -0,0 +1,53 @@
+package com.teragrep.pth10;
```
elliVM, line 1: Missing license

elliVM (changes requested):
Changes look good to me just missing licenses and to apply spotless

elliVM (approved):
LGTM

eemhu:
Added missing SpathEscapedKey in evalStatement, rebased onto main and re-enabled disabled spath tests

elliVM (approved):

src/main/java/com/teragrep/pth10/steps/spath/SpathEscapedKey.java
```diff
@@ -0,0 +1,96 @@
...
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth10.steps.spath;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+public final class SpathEscapedKey implements Serializable {
```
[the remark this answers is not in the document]
kortemik: SpathEscapedKey should have a method unescape that returns SpathKey

kortemik (changes requested):

src/main/java/com/teragrep/pth10/steps/spath/SpathEscapedKey.java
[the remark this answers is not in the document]
Abigael-JT: This has been fixed

src/main/java/com/teragrep/pth10/steps/spath/SpathEscapedKey.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private static final long serialVersionUID = 1L;
+    private final String key;
+
+    public SpathEscapedKey(final String key) {
+        this.key = key;
+    }
+
+    public SpathKey unescaped() {
```
kortemik, line 60: I was thinking this during the day ans thought that is this actually an escaped thing or is it rather a backtick quoted string and usual unquote functionality should.be used? @eemhu Any comments as you did previous quotation fixes

eemhu: I suppose strictly speaking it is more of a backtick quotation but logically it is used to escape `.` (dots), so they can be used in the column names.

kortemik: I think we should make the object re-usable for such purposes and use the generic quotation object that allows specifying the quotation character. this way it would be able to handle 
```
' " `
```
 and possibly other characters as well. current substring way does not handle properly escapes in a string. i.e. 
```
`foo\`bar`
```
 should be unquoted as 
```
foo`bar
```

eemhu: Added backtick support for existing UnquotedText object and also added QuotedText object.

src/main/java/com/teragrep/pth10/ast/UnquotedText.java
```diff
@@ -53,7 +54,7 @@
  */
 public class UnquotedText implements Text {
```
kortemik, line 55: i think this should take the quote character as an argument instead of using regex to match it?

kortemik (changes requested):
i would add a constructor parameter instead of using regex because regex is not very readable.

otherwise looks very clean and usable

src/main/java/com/teragrep/pth10/ast/UnquotedText.java
eemhu: Would it be okay if no quote character is provided it would remove all, and if it is provided it only removes the specific one? I think currently UnquotedText is used to remove all on purpose

kortemik (approved):

kortemik:
good code: merging!

PR #370, opened 2024-10-09
Add exception handling for bloom option json reading

elliVM:
- Catch exceptions thrown by gson when reading filter fields option string to JSON
- Use FilterField object to store filter field options `expected` and `fpp` and test
- Get BloomFilter `bitSize` from FilterField object
- Add tests for incorrect JSON fromat in options and duplicate filter field values

### DIFF
--- a/src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
+++ b/src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
@@ -153,7 +153,8 @@ public final class TeragrepBloomStep extends AbstractStep {
      * @return Dataset unmodified
      */
     private Dataset<Row> createBloomFilter(Dataset<Row> dataset) {
-        writeFilterTypes(this.zeppelinConfig);
+        createFilterTypeTable();
+        writeFilterTypes();
         final BloomFilterTable table = new BloomFilterTable(zeppelinConfig);
         table.create();
         dataset.foreachPartition(new BloomFilterForeachPartitionFunction(this.zeppelinConfig));
@@ -167,7 +168,8 @@ public final class TeragrepBloomStep extends AbstractStep {
      * @return Dataset unmodified
      */
     private Dataset<Row> updateBloomFilter(Dataset<Row> dataset) {
-        writeFilterTypes(this.zeppelinConfig);
+        createFilterTypeTable();
+        writeFilterTypes();
         final BloomFilterTable table = new BloomFilterTable(zeppelinConfig);
         table.create();
         dataset.foreachPartition(new BloomFilterForeachPartitionFunction(this.zeppelinConfig, true));
@@ -191,7 +193,7 @@ public final class TeragrepBloomStep extends AbstractStep {
         return dataset.groupBy("partition").agg(agg.toColumn().as("bloomfilter"));
     }
 
-    private void writeFilterTypes(final Config config) {
+    private void writeFilterTypes() {
         final List<FilterField> fieldList = filterTypes.fieldList();
         final String pattern = filterTypes.pattern();
         final Connection conn = connection.get();
@@ -224,6 +226,23 @@ public final class TeragrepBloomStep extends AbstractStep {
                 }
                 throw new RuntimeException(e);
             }
+        }
+    }
+
+    private void createFilterTypeTable() {
+        // from pth-06/database/bloomdb
+        final String sql = "CREATE TABLE IF NOT EXISTS `filtertype` ("
+                + "    `id`               bigint(20) UNSIGNED   NOT NULL AUTO_INCREMENT PRIMARY KEY,"
+                + "    `expectedElements` bigint(20) UNSIGNED   NOT NULL,"
+                + "    `targetFpp`        DOUBLE(2, 2) UNSIGNED NOT NULL,"
+                + "    `pattern`          VARCHAR(2048)         NOT NULL,"
+                + "    UNIQUE KEY (`expectedElements`, `targetFpp`, `pattern`)" + ") ENGINE = InnoDB"
+                + "  DEFAULT CHARSET = utf8mb4" + "  COLLATE = utf8mb4_unicode_ci;";
+        try (final PreparedStatement statement = connection.get().prepareStatement(sql)) {
+            statement.execute();
+        }
+        catch (SQLException e) {
+            throw new RuntimeException("Error creating `filtertype` table: " + e.getMessage());
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
```diff
@@ -162,32 +158,35 @@ private Dataset<Row> estimateSize(Dataset<Row> dataset) {
     }
 
     public Dataset<Row> aggregate(Dataset<Row> dataset) {
-
-        FilterTypes filterTypes = new FilterTypes(this.zeppelinConfig);
-
-        BloomFilterAggregator agg = new BloomFilterAggregator(inputCol, estimateCol, filterTypes.sortedMap());
-
+        final FilterTypes filterTypes = new FilterTypes(this.zeppelinConfig);
```
51-code, line 161: I would have this FilterTypes object as a class variable rather than the zeppelinConfig. Realistically though, if thinking with single responsibility principle in mind, this object should be four objects. Now it's resposnsible for creating, updating, estimating and aggregating (like in the Enum). If it was refactored into multiple objects, the FilterTypes as class variable suggestion would make more sense too. If that sounds good, I would open a new issue and eventually another PR for that so this one doesn't get too big.

src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
```diff
@@ -162,32 +158,35 @@ private Dataset<Row> estimateSize(Dataset<Row> dataset) {
...
+        final SortedMap<Long, Double> map = new TreeMap<>();
+        final List<FilterField> fieldList = filterTypes.fieldList();
+        for (final FilterField field : fieldList) {
+            map.put(field.expected(), field.fpp());
+        }
+        final BloomFilterAggregator agg = new BloomFilterAggregator(inputCol, estimateCol, map);
```
51-code, line 167: I like the idea of having the FilterTypes as a list but it feels weird that then it has to be put into a map instantly anyways like before. Could the BloomFilterAggregator be refactored to work with the list too?

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/FilterField.java
```diff
@@ -0,0 +1,94 @@
...
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth10.steps.teragrep.bloomfilter;
+
+import org.apache.spark.util.sketch.BloomFilter;
+
+import java.util.Objects;
+
+public class FilterField {
```
51-code, line 52: Final keyword missing in class declaration.

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/FilterTypes.java
```diff
@@ -73,35 +72,33 @@ public FilterTypes(Config config) {
...
-                throw new RuntimeException("JSON did not have expected values of 'expected' or 'fpp'");
-            }
+    public List<FilterField> fieldList() {
+        final List<FilterField> fieldList;
+        try {
+            final Gson gson = new Gson();
+            fieldList = gson.fromJson(sizesJsonString(), new TypeToken<List<FilterField>>() {
+            }.getType());
+        }
+        catch (Exception e) {
```
51-code, line 82: Would it be possible to catch a more specific exception, like JsonSyntaxException (I'm not 100% sure what this fromJson() throws)

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/FilterTypes.java
```diff
@@ -73,35 +72,33 @@ public FilterTypes(Config config) {
...
+            throw new RuntimeException("Found duplicate values in 'dpl.pth_06.bloom.db.fields'");
         }
-        return sizesMapFromJson;
+        return fieldList;
     }
 
     public Map<Long, Long> bitSizeMap() {
-        final Map<Long, Double> filterSizes = sortedMap();
+        final List<FilterField> fieldList = fieldList();
+        System.out.println(fieldList);
```
51-code, line 97: A debug print is left in the code here.

51-code (changes requested):
Requested some changes.

src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
elliVM: Good suggestion I am looking into it but I also created and issue about splitting this object https://github.com/teragrep/pth_10/issues/374

src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
elliVM: Indeed a bit unfortunate result of refactoring but refactored BloomFilterAggregator would require new release of dpf_03. Also would couple dpf_03 to this custom object type a general map might be better for use elsewhere.

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/FilterTypes.java
elliVM: catching exceptions `JsonIOException` and `JsonSyntaxException` thrown by `fromJson()` now

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/FilterTypes.java
elliVM: removed!

src/main/java/com/teragrep/pth10/steps/teragrep/bloomfilter/FilterField.java
elliVM: made final

src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepBloomStep.java
elliVM: Moved FilterTypes and LazyConnection to class variables, number of class variables are high for now but refactoring is easier with this change. 
Refactoring will also lower the number of variables of this class.

51-code (approved):
LGTM now with the fixes.

eemhu (changes requested):
rebase needed

elliVM:
Will do a new PR as this one has diverged too much

PR #372, opened 2024-10-10
Fix hardcoded database name in constraint

elliVM:
## Get journaldb name from config to SQL constraint
- FilterTypes gets journaldbName method
- TableSQL uses journalDBName when creating table SQL
- Test for custom journalDBName


eemhu (approved):
looks good to me

kortemik (approved):

Issue #384, opened 2024-10-22
add backing support for | teragrep exec foreachBatch


kortemik:
**Description**
see https://github.com/teragrep/pth_03/issues/83

Issue #134, opened 2023-11-14
 Tests only create one batch of data


kortemik:
ID <530>
Reporter <@JohnJacobGuy>

All of the tests seem to create one batch of data. In diffrent enviroment, most if not all datasets will generate multiple batches. This discrepancy is not ideal since the tests should closely emulate the actual deployment environments as closely as possible.

Seemed to happen even before the change to reading JSON files as streams, using the RowFactory and RowMemoryStream method.

#396

PR #390, opened 2024-10-24
#384: Add `| teragrep exec foreachbatch` command that sets processing to batch mode

eemhu:
Adds `| teragrep exec foreachbatch` command, which turns off batch collect and sets processing to batch mode via step properties.

See issue
* #384

### DIFF
--- a/src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepForEachBatchStep.java
+++ b/src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepForEachBatchStep.java
@@ -50,7 +50,7 @@ import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.streaming.StreamingQueryException;
 
-public class TeragrepForEachBatchStep extends AbstractStep {
+public final class TeragrepForEachBatchStep extends AbstractStep {
 
     public TeragrepForEachBatchStep() {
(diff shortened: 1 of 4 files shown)


src/main/java/com/teragrep/pth10/steps/teragrep/TeragrepForEachBatchStep.java
```diff
@@ -0,0 +1,64 @@
...
+ * a licensee so wish it.
+ */
+package com.teragrep.pth10.steps.teragrep;
+
+import com.teragrep.pth10.steps.AbstractStep;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.streaming.StreamingQueryException;
+
+public class TeragrepForEachBatchStep extends AbstractStep {
```
51-code, line 53: Class can be final.

51-code (changes requested):
I guess this has to be tested in QA mostly as the unit tests don't support batches of data?

eemhu:
> I guess this has to be tested in QA mostly as the unit tests don't support batches of data?

Yes, with the current testing setup it's not really possible. As-is the test only really tests that the command produces expected results data-wise, and with only one batch.

51-code (approved):
LGTM now, I don't believe testing batches is possible in unit tests. There's an open issue #134 about that.

kortemik (approved):

Issue #326, opened 2024-09-16
Implement endtimeu command


Abigael-JT:
**Description**

 endtimeu doesn't work in dpl
**Use case or motivation behind the feature request**

 endtimeu should take unix timestamp as a parameter and have it as end time
**Related issues**

 #324 
**Additional context**

Issue #325, opened 2024-09-16
Implement starttimeu command


Abigael-JT:
**Description**

 starttimeu doesn't work in dpl
**Use case or motivation behind the feature request**

starttimeu should take unix timestamp as a parameter and have it as a start time
**Related issues**

 #324 
**Additional context**

Issue #324, opened 2024-09-13
Relativetime Endtimeu and Starttimeu tests


Abigael-JT:
**Description**

These tests were disabled because starttimeu and endtimeu were not implemented

The status of this will be looked at in issue https://github.com/teragrep/pth_10/issues/295

PR #399, opened 2024-10-30
#325, #326: Starttimeu and endttimeu implementation

eemhu:
Adds 
* EpochTimestamp
* TimeQualifier objects.
* Starttimeu, endttimeu support.
* tests

See issues for more information
* #325 
* #326 
* #324

### DIFF
--- a/src/main/java/com/teragrep/pth10/ast/commands/logicalstatement/TimeStatement.java
+++ b/src/main/java/com/teragrep/pth10/ast/commands/logicalstatement/TimeStatement.java
@@ -199,7 +199,7 @@ public class TimeStatement extends DPLParserBaseVisitor<Node> {
         String value = ctx.getChild(1).getText();
 
         
-        TimeQualifier tq = new TimeQualifier(value, catCtx.getTimeFormatString(), node.getSymbol(), doc);
+        TimeQualifier tq = new TimeQualifier(value, catCtx.getTimeFormatString(), node.getSymbol().getType(), doc);
 
         if (tq.isStartTime()) {
             startTime = tq.epoch();
@@ -224,7 +224,7 @@ public class TimeStatement extends DPLParserBaseVisitor<Node> {
         TerminalNode node = (TerminalNode) ctx.getChild(0);
         String value = ctx.getChild(1).getText();
 
-        TimeQualifier tq = new TimeQualifier(value, catCtx.getTimeFormatString(), node.getSymbol(), doc);
+        TimeQualifier tq = new TimeQualifier(value, catCtx.getTimeFormatString(), node.getSymbol().getType(), doc);
 
         if (tq.isStartTime()) {
             startTime = tq.epoch();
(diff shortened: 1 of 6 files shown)


src/main/java/com/teragrep/pth10/ast/time/EpochTimestamp.java
```diff
@@ -0,0 +1,69 @@
...
+            rv = epochFromString(value, timeformat);
+        }
+
+        return rv;
+    }
+
+    // Uses defaultTimeFormat if timeformat is null and DPLTimeFormat if timeformat isn't null (which means that the
+    // timeformat= option was used).
+    private long epochFromString(String value, String timeFormatString) {
+        value = new UnquotedText(new TextString(value)).read(); // erase the possible outer quotes
```
elliVM, line 37: Could this rather be new local variable instead of reassigning `value`?

src/main/java/com/teragrep/pth10/ast/time/EpochTimestamp.java
```diff
@@ -0,0 +1,69 @@
...
+        long timevalue = 0;
+        if (timeFormatString == null || timeFormatString.equals("")) {
```
elliVM, line 39: could use `.isEmpty()` instead of equals check here

elliVM (changes requested):
left some comments and  `mvn spotless:check` reported some issues

src/main/java/com/teragrep/pth10/ast/time/EpochTimestamp.java
eemhu: fixed

src/main/java/com/teragrep/pth10/ast/time/EpochTimestamp.java
eemhu: fixed

elliVM (approved):
Changes look good

src/main/java/com/teragrep/pth10/ast/commands/logicalstatement/TimeStatement.java
```diff
@@ -191,65 +183,24 @@ public Node visitTimeQualifier(DPLParser.TimeQualifierContext ctx) {
...
-        RelativeTimeParser rtParser = new RelativeTimeParser();
-        // Is time given as absolute
-        // Try to check if it is relative and catch exception
-        try {
-            // relative time
-            RelativeTimestamp rtTimestamp = rtParser.parse(value); // might throw NFE if not relative timestamp
-            timevalue = rtTimestamp.calculate(now);
+        String value = ctx.getChild(1).getText();
+
+        TimeQualifier tq = new TimeQualifier(value, catCtx.getTimeFormatString(), node.getSymbol().getType(), doc);
```
kortemik, line 191: i think the syntax supports specifying timeformat multiple times, here it is on the catCtx indicating only one timeformat would be supported for the whole query, however

```
index=search_bench timeformat=%s earliest=1 timeformat=%Y-%m-%dT%H:%M:%S latest=2024-11-14T08:00:00
```

should be a valid query as well.

perhaps it is a worth of another issue?

kortemik (approved):

PR #400, opened 2024-10-30
move tokenizer step to teragrep package and rename

elliVM:
- move tokenizer step to teragrep package
- rename to TeragrepTokenizerStep
- remove unused interface and logger


51-code (approved):
LGTM. I'll approve, but two things:

1. TokenizerStep was renamed to TeragrepTokenizerStep but the AbstractTokenizerStep wasn't renamed in the same fashion, mistake or not? Either way the naming seems good to me.
2. TokenizerStep should be refactored to multiple objects based on the enum tokenizerFormat, new issue about that.

PR #404, opened 2024-11-04
replace explode with explode_outer

elliVM:
change bloom estimate to use `explode_outer` 

- `explode()` returns only non empty results so rows are lost
- `explode_outer()` returns null that `approxCountDistinct()` can calculate to return 0


51-code (approved):
LGTM. Clarification maybe needed: does the test produce an empty tokenizer column because the _raw value is empty in index_Empty?

kortemik (approved):

Issue #439, opened 2024-12-02
unexpected results from aggregation earliest() / latest() on _raw field


Abigael-JT:
**Describe the bug**

 This query fails:
index=crud earliest=-3y

| stats earliest(_raw), latest(_raw)

**Expected behavior**

 Should return the earliest and latest event in _raw field
 
But Returns:
Job aborted due to stage failure
....

**How to reproduce**

 Run a query similar to the one above

**Screenshots**

 
**Software version**

pth_10 version: 8.2.1
pth_03 version: 9.2.0
 
**Desktop (please complete the following information if relevant):**
 - OS: 
 - Browser: 
 - Version: 
 
**Additional context**

Abigael-JT:
fix issue #439 to make it easier to troubleshoot

PR #444, opened 2024-12-04
#439: Only retain current earliest or latest row instead of all

eemhu:
Fixes issue
* #439 

It was happening with large datasets, most likely cause being all rows were saved into the buffer. In this PR, only the currently earliest / latest row is saved to the buffer. Also the buffer is now immutable and merging logic was moved to the aggregator.

### DIFF
--- a/src/main/java/com/teragrep/pth10/ast/commands/aggregate/UDAFs/CurrentTimestampImpl.java
+++ b/src/main/java/com/teragrep/pth10/ast/commands/aggregate/UDAFs/CurrentTimestampImpl.java
@@ -45,55 +45,39 @@
  */
 package com.teragrep.pth10.ast.commands.aggregate.UDAFs;
 
-import com.teragrep.pth10.ast.commands.aggregate.UDAFs.BufferClasses.EarliestLatestBuffer;
-import org.apache.spark.sql.Encoder;
-import org.apache.spark.sql.Encoders;
+import java.sql.Timestamp;
 
-import java.io.Serializable;
+public final class CurrentTimestampImpl implements CurrentTimestamp {
 
-/**
- * Used for rate() function
- */
-public class EarliestLatestAggregator_Double extends EarliestLatestAggregator<Double> implements Serializable {
+    private final Timestamp timestamp;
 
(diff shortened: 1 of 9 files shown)


src/test/java/com/teragrep/pth10/statsTransformationTest.java
```diff
@@ -190,7 +190,6 @@ void statsTransform_AggRate_Test() {
     void statsTransform_AggEarliest_Test() {
```
51-code, line 190: Can we have a test for the `| stats earliest(_raw), latest(_raw)` combo?

src/main/java/com/teragrep/pth10/ast/commands/aggregate/UDAFs/CurrentTimestampImpl.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth10.ast.commands.aggregate.UDAFs;
+
+import java.sql.Timestamp;
+
+public final class CurrentTimestampImpl implements CurrentTimestamp {
```
51-code, line 50: Unit tests for CurrentTimestamp? It's quite trivial but isBefore and isAfter would benefit at least and equality tests would be nice.

51-code (changes requested):
The code looks good to me but I left some comments about unit testing.

src/test/java/com/teragrep/pth10/statsTransformationTest.java
eemhu: added

src/main/java/com/teragrep/pth10/ast/commands/aggregate/UDAFs/CurrentTimestampImpl.java
eemhu: added

51-code (approved):
LGTM

kortemik (approved):

kortemik (approved):
looks good!

PR #445, opened 2024-12-04
Fix bloom translation tests

### LINKED ISSUE #443: Tests failures in newest main branch commit
**Description**

TeragrepTest.testTeragrepBloomCreateTranslation
TeragrepTest.testTeragrepBloomUpdateTranslation

fails

```
java.lang.IllegalArgumentException: Missing table parameter in '| teragrep exec bloom'.
```

51-code:
Fixes #443 .

Bloom translation tests did not include the table and regex parameters that were moved from spark config to dpl language earlier, which caused two tests to fail. Added the missing parameters to the test queries.

Would like to improve these tests in the future, currently they just test that the bloom mode is correct. Making all fields public for translationTests (like elsewhere) doesn't seem too smart, so overriding equals and hashcode methods would be the best option. Creating equality tests is another challenge as steps aren't immutable currently because of the AbstractStep class. So, a refactoring should be done for Steps. I'll make some issues about these problems.

### DIFF
--- a/src/test/java/com/teragrep/pth10/translationTests/TeragrepTest.java
+++ b/src/test/java/com/teragrep/pth10/translationTests/TeragrepTest.java
@@ -326,7 +326,7 @@ public class TeragrepTest {
 
     @Test
     void testTeragrepBloomCreateTranslation() {
-        final String query = "| teragrep exec bloom create";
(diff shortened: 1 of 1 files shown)


eemhu (approved):

51-code:
Opened issue #446  about improving the translation tests in general.

PR #429, opened 2024-11-21
issue-65: Improve test assertions 

Abigael-JT:
Updated expectedSchema in assertions to be created from java objects

Updated schema assertions to use java objects

Fixes for:
* #65 

Related PR:
* #447

Issue #65, opened 2023-11-14
Improve test assertations 


kortemik:
ID <453>
Reporter <@JohnJacobGuy>

Currently test cases use string comparisation with hard code expected strings. 

Changes to these are hard to track and implement.

They must be changed at least so that the expected strings are created from java objects instead of being hardcoded.

i.e. in following

e = "StructType(StructField(_raw,StringType,true), StructField(_time,StringType,true), StructField(host,StringType,true), StructField(index,StringType,true), StructField(offset,LongType,true), StructField(origin,StringType,true), StructField(partition,StringType,true), StructField(source,StringType,true), StructField(sourcetype,StringType,true))";

String resSchema=res.schema().toString();
assertEquals(e, resSchema);

must become

StructType expectedSchema = new StructType(
        new StructField[]{
            new StructField("_time", DataTypes.TimestampType, false, new MetadataBuilder().build()),
            new StructField("_raw", DataTypes.StringType, false, new MetadataBuilder().build()),
            new StructField("index", DataTypes.StringType, false, new MetadataBuilder().build()),
            new StructField("sourcetype", DataTypes.StringType, false, new MetadataBuilder().build()),
            new StructField("host", DataTypes.StringType, false, new MetadataBuilder().build()),
            new StructField("source", DataTypes.StringType, false, new MetadataBuilder().build()),
            new StructField("partition", DataTypes.StringType, false, new MetadataBuilder().build()),
            new StructField("offset", DataTypes.LongType, false, new MetadataBuilder().build()),
            new StructField("origin", DataTypes.StringType, false, new MetadataBuilder().build()),
        }
    );
String resSchema=res.schema().toString();
assertEquals(expectedSchema.toString(), resSchema);

#163

PR #447, opened 2024-12-05
updated schema assertions to use java objects

Abigael-JT:
Updated schema assertions to use java objects

Fixes for:
* #65 

Related PR:
* #429


kortemik (approved):
Good improvement! Approving

Issue #435, opened 2024-11-26
logicalStatement latest does not apply correctly with second fractions


kortemik:
**Describe the bug**
logicalStatement latest does not apply correctly with second fractions
 
**Expected behavior**
latest with subsecond increment should include data
 
**How to reproduce**
```
%dpl

index=crud_small earliest=2021-12-25T16:24:39.000+02:00 latest=2021-12-25T20:52:57.000+02:00
```

results into two rows as should
```
_time	_raw
2021-12-25T20:52:56.000+02:00	{"timestamp_epoch": 1640458376, "timestamp": "2021-12-25T20:52:56+02:00", "ipaddress": "127.0.6.209", "operation": "update", "count": 19, "elapsed": 7.36, "balance": -511.067, "success": true}
2021-12-25T16:24:39.000+02:00	{"timestamp_epoch": 1640442279, "timestamp": "2021-12-25T16:24:39+02:00", "ipaddress": "127.0.1.169", "operation": "delete", "count": -78, "elapsed": 3.21, "balance": -790.894, "success": true}
```

while

```
%dpl

index=crud_small earliest=2021-12-25T16:24:39.000+02:00 latest=2021-12-25T20:52:56.001+02:00
```

results only into one row
```
_time	_raw
2021-12-25T16:24:39.000+02:00	{"timestamp_epoch": 1640442279, "timestamp": "2021-12-25T16:24:39+02:00", "ipaddress": "127.0.1.169", "operation": "delete", "count": -78, "elapsed": 3.21, "balance": -790.894, "success": true}
```

 
**Screenshots**

 
**Software version**
pth_10 8.2.0
 
**Desktop (please complete the following information if relevant):**
 - OS: 
 - Browser: 
 - Version: 
 
**Additional context**

PR #454, opened 2024-12-17
#435: logicalStatement latest does not apply correctly with second fractions

eemhu:
Fixes issue:
* #435 

Changes:
* Rounds up to the nearest second if fractions-of-second are used in the latest timeQualifier. Earliest timeQualifier ignores subseconds / rounds them down.
* Added timezones to test data.

### DIFF
--- a/src/main/java/com/teragrep/pth10/ast/DefaultTimeFormat.java
+++ b/src/main/java/com/teragrep/pth10/ast/DefaultTimeFormat.java
@@ -56,6 +56,22 @@ import java.util.Date;
  */
 public class DefaultTimeFormat {
 
+    private final String[] formats;
+
+    public DefaultTimeFormat() {
+        this(new String[] {
+                "MM/dd/yyyy:HH:mm:ss",
+                "yyyy-MM-dd'T'HH:mm:ss.SSSXXX",
+                "yyyy-MM-dd'T'HH:mm:ss.SSS",
+                "yyyy-MM-dd'T'HH:mm:ssXXX",
+                "yyyy-MM-dd'T'HH:mm:ss"
+        });
+    }
+
+    public DefaultTimeFormat(String[] formats) {
+        this.formats = formats;
+    }
+
     /**
      * Calculate the epoch from given string.
      * 
@@ -73,40 +89,15 @@ public class DefaultTimeFormat {
      * @return Date parsed from the given string
      */
     public Date parse(String time) {
-        // Try parsing all the three default timeformats
-        Date date;
-
-        int attempt = 0;
-        while (true) {
+        // Try parsing all provided time formats in order
+        for (final String format : formats) {
             try {
-                if (attempt == 0) {
-                    // Use default format (MM/dd/yyyy:HH:mm:ss)
-                    // Use system default timezone
-                    date = this.parseDate(time, "MM/dd/yyyy:HH:mm:ss");
-                }
-                else if (attempt == 1) {
-                    // On first fail, try ISO 8601 with timezone offset, e.g. '2011-12-03T10:15:30+01:00'
-                    date = this.parseDate(time, "yyyy-MM-dd'T'HH:mm:ssXXX");
-                }
-                else {
-                    // On second fail, try ISO 8601 without offset, e.g. '2011-12-03T10:15:30'
-                    // Use system default timezone
-                    date = this.parseDate(time, "yyyy-MM-dd'T'HH:mm:ss");
-                }
-                break;
-
+                return parseDate(time, format);
             }
-            catch (ParseException e) {
-                if (attempt > 1) {
-                    throw new RuntimeException("TimeQualifier conversion error: <" + time + "> can't be parsed.");
-                }
-            }
-            finally {
-                attempt++;
+            catch (ParseException ignored) {
             }
         }
-
-        return date;
+        throw new RuntimeException("TimeQualifier conversion error: <" + time + "> can't be parsed.");
     }
 
     private Date parseDate(String time, String timeFormat) throws ParseException {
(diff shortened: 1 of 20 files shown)


src/main/java/com/teragrep/pth10/ast/DefaultTimeFormat.java
```diff
@@ -112,6 +131,13 @@ else if (attempt == 1) {
...
-        return sdf.parse(time);
+        Date date = sdf.parse(time);
+
+        // If date is for latest timeQualifier and has fractions-of-second, add 1 second to capture events
+        // that are on the same second
+        if (date.toInstant().getNano() > 0 && isLatest) {
+            date = Date.from(date.toInstant().plus(1, ChronoUnit.SECONDS));
+        }
+        return date;
     }
```
elliVM, line 142: I had couple of ideas for this logic and I came up with this. I think this could just be replaced by having a list of formats and a for loop that tries to parse each format and returns the date if no exception was thrown. if the end of the for loop is reached an exception is thrown. this would also leave room for easily adding or removing formats in the future. List of formats could be given in the constructor and a secondary constructor could have a default list of formats. Here is an example of code

```
public Date parse(final String time) {
        // Try parsing all formats
        for (String format: formats) {
            try {
                return parseDate(time, format);
            } catch (ParseException ignored) {
            }
        }
        throw new RuntimeException("Could not parse time <" + time + ">");
    } 
```
you could keep track of the parser error and return the latest one if you think it is important information

src/test/java/com/teragrep/pth10/DefaultTimeFormatTest.java
```diff
@@ -61,6 +61,24 @@ void enforceTimeZone() {
...
+
+    @Test
+    void ISO8601_fractions_earliest_Test() {
+        String time = "2023-12-18T12:40:53.001+03:00"; // different timeformat than default (default would be +02:00)
+        long expected = 1702892453L;
+        long actual = new DefaultTimeFormat(false).getEpoch(time);
+
+        Assertions.assertEquals(expected, actual);
+    }
+
```
elliVM, line 81: Could you also add a test that a unparseable time throws an exception with a expected message?

elliVM (commented):
Left some comments for your consideration, re-request review if you think they are not necessary or have made some changes.

src/test/java/com/teragrep/pth10/DefaultTimeFormatTest.java
eemhu: added

src/main/java/com/teragrep/pth10/ast/DefaultTimeFormat.java
eemhu: fixed

elliVM (approved):
looks good

kortemik:
Nice enhancement. Would it be more proper to make earliest/latest value by composition instead of using constructor argument of true/false? I got distracted by it during the review and think that it could use some more care.

eemhu:
> Nice enhancement. Would it be more proper to make earliest/latest value by composition instead of using constructor argument of true/false? I got distracted by it during the review and think that it could use some more care.

Changed DefaultTimeFormat, DPLTimeFormat and EpochTimestamp (now called InstantTimestamp) to return Instant rather than unix epoch. Now when latest is used in TimeQualifier, InstantTimestamp will be decorated with RoundedUpTimestamp to perform the rounding up functionality.

src/main/java/com/teragrep/pth10/ast/time/RelativeTimestamp.java
kortemik, line 1: Please create an issue about refactoring RelativeTimestamp or fix it. It's not even null free.

src/main/java/com/teragrep/pth10/ast/time/InstantTimestamp.java
```diff
@@ -52,46 +52,47 @@
...
         }
         else {
             // TODO: should be included in DPLTimeFormat
             if (timeFormatString.equals("%s")) {
-                return Long.parseLong(unquotedValue);
+                return Instant.ofEpochSecond(Long.parseLong(unquotedValue));
             }
             try {
-                timevalue = new DPLTimeFormat(timeFormatString).getEpoch(unquotedValue);
+                timevalue = new DPLTimeFormat(timeFormatString).instantOf(unquotedValue);
```
kortemik, line 95: composition would be preferred approach instead of converter methods

src/main/java/com/teragrep/pth10/ast/time/RoundedUpTimestamp.java
kortemik, line 1: This is only used for Archive datasource, right?

src/main/java/com/teragrep/pth10/ast/time/RelativeTimestamp.java
eemhu: https://github.com/teragrep/pth_10/issues/476

src/main/java/com/teragrep/pth10/ast/time/RoundedUpTimestamp.java
eemhu: It is currently used for both XML queries and spark columns as they both use unix epoch (seconds) so it would not filter out the fractions of seconds

src/main/java/com/teragrep/pth10/ast/time/InstantTimestamp.java
eemhu: will make a new issue for this

eemhu:
rebased

kortemik (approved):

kortemik:
good work @eemhu & @Abigael-JT

PR #469, opened 2025-01-14
fix coverity reported resource leak and logging in TeragrepSystemStep

elliVM:
- Use resource with try-with-resources
- Guard logger method calls


51-code (approved):
LGTM

kortemik (approved):

PR #470, opened 2025-01-14
Update ci and coverity workflow

elliVM:
Split CI and Coverity into separate workflows. CI.yaml has `Install test dependencies` step that is carried over from old workflow otherwise same as in example workflow.
 
- ci.yaml runs testing
- coverity.yaml runs coverity uploading


.github/workflows/coverity.yaml
```diff
@@ -0,0 +1,99 @@
...
+              echo 'Waiting for 15 seconds and retrying'
+              sleep 15
+            else
+              break
+            fi
+          done
+
+      - name: Compile Coverity
+        run: |
+          ${RUNNER_TEMP}/${{ env.COVERITY }}/bin/cov-build --dir ${RUNNER_TEMP}/cov-int mvn -B -Drevision=${{ env.RELEASE_VERSION }} -Dsha1= -Dchangelist= clean verify
```
StrongestNumber9, line 79: Looks otherwise valid but the coverity.yaml will also require those external dependencies as it runs normal mvn execution. Add it and this should be good to go

StrongestNumber9 (changes requested):

.github/workflows/coverity.yaml
elliVM: added the external test dependencies to coverity.yaml

StrongestNumber9 (approved):
Seems good to me

PR #475, opened 2025-01-21
throw exception if teragrep datasource is queried when its disabled in config

elliVM:
Throws a runtime exception in case trying to construct an archive query when option `dpl.pth_06.enabled=false`.

### DIFF
--- a/src/main/java/com/teragrep/pth10/datasources/DPLDatasource.java
+++ b/src/main/java/com/teragrep/pth10/datasources/DPLDatasource.java
@@ -84,16 +84,11 @@ public class DPLDatasource {
     public Dataset<Row> constructStreams(ArchiveQuery archiveQuery, boolean isMetadataQuery) {
         // resolve archive Query which is then used with archiveDatasource
         LOGGER.info("DPL Interpreter ArchiveQuery=<[{}]>", archiveQuery);
-
-        Dataset<Row> archiveDS = null;
         LOGGER.info("DPL Interpreter constructStream config=<[{}]>", config);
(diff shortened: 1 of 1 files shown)


src/main/java/com/teragrep/pth10/datasources/DPLDatasource.java
```diff
@@ -84,16 +84,11 @@ public DPLDatasource(DPLParserCatalystContext catCtx) {
...
         LOGGER.info("DPL Interpreter ArchiveQuery=<[{}]>", archiveQuery);
-
-        Dataset<Row> archiveDS = null;
         LOGGER.info("DPL Interpreter constructStream config=<[{}]>", config);
-        // table name per paragraph
-        if (config.getBoolean("dpl.pth_06.enabled")) {
-            // Execute search against ArchiveDataSource
-            archiveDS = archiveStreamConsumerDataset(archiveQuery, isMetadataQuery);
+        if (!config.getBoolean("dpl.pth_06.enabled")) {
+            throw new RuntimeException("Teragrep datasource was disabled: <dpl.pth_06.disabled=false>");
```
kortemik, line 89: please check the exception content, it is inversely named.

kortemik (changes requested):

src/main/java/com/teragrep/pth10/datasources/DPLDatasource.java
elliVM: fixed exception message

kortemik (approved):

PR #479, opened 2025-01-23
remove null assignments and check that the map is not empty check

elliVM:
- Check that map is not empty before iterating it
-  Remove the use of null

### DIFF
--- a/src/test/java/com/teragrep/pth10/statsTransformationTest.java
+++ b/src/test/java/com/teragrep/pth10/statsTransformationTest.java
@@ -45,7 +45,6 @@
  */
 package com.teragrep.pth10;
 
-import com.teragrep.pth10.ast.NullValue;
 import org.apache.spark.sql.Row;
 import org.junit.jupiter.api.*;
 import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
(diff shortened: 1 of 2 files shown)


51-code (changes requested):
I believe the logic has changed here somewhat. The map might be empty if there is no data found with the given query. The old version with the null might have let the query pass with an empty dataset, but the refactored version throws an exception. Letting the query through is probably the better option! Then the user would know that there was no data in the given time frame, because an empty dataset would be returned in the UI.

There is no test for this right now. There is one test for mode() in statsTransformationTest.java, but a test with an empty dataset could be added to see how it behaves currently as well as making sure that the logic doesn't change.

kortemik (approved):

kortemik (changes requested):
see change proposal by @51-code

elliVM:
made some changes that on an empty dataset result will be a empty column. Added test for empty dataset

src/test/java/com/teragrep/pth10/statsTransformationTest.java
```diff
@@ -45,6 +45,7 @@
  */
 package com.teragrep.pth10;
 
+import com.teragrep.pth10.ast.NullValue;
```
51-code, line 48: Seems to be an unused import? Maybe spotless has to be run.

51-code (changes requested):
Now the solution for the mode aggregation looks really good, nice job. Left a comment about an import, maybe it was left behind after rewriting the test.

51-code (approved):
Running spotless fixed the problem I commented on earlier. Approved!

kortemik (approved):

PR #483, opened 2025-01-23
add default exception cases for switches to avoid null column values

elliVM:
- Add exception message to unrecognized method commands to avoid NPE
- Remove null values in methods


51-code (approved):

kortemik (approved):
all good, thank you! merging now

PR #490, opened 2025-01-29
Timechart refactoring

### LINKED ISSUE #473: Refactor timechart command
**Description**

Before working on timechart command the current code should be refactored, i.e. the way TimechartTransformation and TimechartStep collaborate. Classes are using getters and setters, there's abstract classes and so on. TimechartStep should be made mutable etc. This will help later on when creating new functionality.

51-code:
Fixes #473 .

Refactored TimechartStep:

- Made it immutable
- Removed the AbstractTimechartStep class which basically just contained getters and setters and the class variables

Reworked TimechartTest:

- The tests broke because they checked the correct translation of the parse tree with the getters from TimechartStep. Reworked the tests to assert equality with an expected TimechartStep object instead.

Refactored TimechartTransformation:

- Removed unused parameters, unused code, commented out code and so on.
- Refactored the visiting logic. Old version used a for loop to go over the child nodes and collect values for the TimechartStep. The new version calls visitChildren() and collects the values for TimechartStep to class variables instead.


eemhu:
Check the possibility of using `ContextValue` to avoid object mutability.

51-code:
Refactored the Transformation class to be immutable with span and divByInst variables using the ContextValue object.

Refactored parsing span quite a lot as it was heavily dependent on utility functions. Created new objects, SpanParameter and TimeRange, because of it.

AggregateFunctions are still mutable as it's impossible to make it immutable without grammar changes to the renaming of the columns first and there's an open PR about that in PTH-03: [# 94](https://github.com/teragrep/pth_03/pull/94)

eemhu (commented):
I think the new objects introduced could use some unit tests

51-code:
> I think the new objects introduced could use some unit tests, otherwise looks good

Created unit tests for SpanParameter and TimeRange. The contextValue classes are very hard to test because of the ANTLR parse trees involved. The ANTLR parsing is covered by the translation tests anyways, so I think creating tests for just the two objects is sufficient.

eemhu (approved):

PR #505, opened 2025-02-06
Create profile for shaded jar, create rpm and bundle shaded jar there

StrongestNumber9:
As little disruption as possible to maven usage this way


StrongestNumber9:
Draft until pth_07/zep_01 parts are also functional

StrongestNumber9:
The release github workflow isn't tested as it would currently fail on maven central release but it has been loaned from existing and working project

kortemik (approved):

Issue #446, opened 2024-12-04
Improve translation tests


51-code:
**Description**

Translation tests currently use protected or public variables in Step classes to check if the translation from DPL was correct. A better way could be to override equals and hashCode methods, then write tests that compare a carefully crafted step to the translation result.

Not as simple of a task as it sounds though, as equality tests will force the objects to be immutable. Right now, AbstractStep uses a mutable field `aggregationsUsedBefore`, so it has to be refactored. Some regular steps are probably mutable currently as well.

Related to #202 .

PR #513, opened 2025-02-18
Throw an exception if ChartStep gets no aggregation columns 

Abigael-JT:
- Added an exception if ChartStep gets no aggregation columns
- Added unit tests for chart count aggregation with no specified field and  on an empty dataset
- Moved code from AbstractChartStep to ChartStep, and removed it

### DIFF
--- a/src/main/java/com/teragrep/pth_10/steps/chart/ChartStep.java
+++ b/src/main/java/com/teragrep/pth_10/steps/chart/ChartStep.java
@@ -67,10 +67,6 @@ public final class ChartStep extends AbstractStep {
 
     @Override
     public Dataset<Row> get(final Dataset<Row> dataset) {
-        if (dataset == null) {
-            return null;
-        }
-
         if (listOfAggrExpressions.isEmpty()) {
             throw new RuntimeException("ChartStep did not receive the necessary aggregation columns");
         }


51-code (approved):
Approved. There's a few things that might need a rework in the future but i dont think they are a problem in this PR. Just to note for the future reviewers:

- Getters are the way of doing translation tests, and are therefore needed as that is the simplest way for now. There's an open issue #446  about that.
- Without dwelling too deep into DPL grammar, listOfAggrExpressions probably can't ever be empty as it would be invalid DPL. A test throwing the RuntimeException is probably impossible to make with DPL queries.

src/main/java/com/teragrep/pth10/steps/chart/ChartStep.java
```diff
@@ -53,29 +54,41 @@
...
         // skip first one as .agg has strange arguments
-        Seq<Column> seqOfExpr = JavaConversions.asScalaBuffer(listOfExpr.subList(1, listOfExpr.size()));
-        Seq<Column> seqOfGroupBy = JavaConversions.asScalaBuffer(listOfGroupBy);
+        final Seq<Column> seqOfExpr = JavaConversions.asScalaBuffer(listOfAggrExpressions.subList(1, listOfAggrExpressions.size()));
+        final Seq<Column> seqOfGroupBy = JavaConversions.asScalaBuffer(groupByList);
 
         return dataset.groupBy(seqOfGroupBy).agg(mainExpr, seqOfExpr);
     }
+
+    public List<Column> getListOfAggrExpressions() {
```
eemhu, line 87: get prefix could be removed

Abigael-JT: removed

eemhu (approved):

eemhu (changes requested):
please run spotless

elliVM:
Rebased and ran spotless

src/main/java/com/teragrep/pth_10/steps/chart/ChartStep.java
```diff
@@ -53,29 +54,42 @@
...
+    public ChartStep(final List<Column> listOfAggrExpressions, final List<Column> groupByList) {
+        this.listOfAggrExpressions = listOfAggrExpressions;
+        this.groupByList = groupByList;
         this.properties.add(CommandProperty.AGGREGATE);
     }
 
     @Override
-    public Dataset<Row> get(Dataset<Row> dataset) {
+    public Dataset<Row> get(final Dataset<Row> dataset) {
         if (dataset == null) {
```
eemhu, line 70: this null check could probably be removed also

PR #519, opened 2025-03-05
Issue_458: Improved error message for incorrect timestamp format

Abigael-JT:
- Improved error message for incorrect timestamp format
- Added a debug log message for unparseable relative timestamp

### DIFF
--- a/src/main/java/com/teragrep/pth10/ast/DefaultTimeFormat.java
+++ b/src/main/java/com/teragrep/pth10/ast/DefaultTimeFormat.java
@@ -75,30 +75,35 @@ public class DefaultTimeFormat {
     public Date parse(String time) {
         // Try parsing all the three default timeformats
         Date date;
-
+        final String defaultFormat = "MM/dd/yyyy:HH:mm:ss";
+        final String withTimezone = "yyyy-MM-dd'T'HH:mm:ssXXX";
+        final String withoutTimezone = "yyyy-MM-dd'T'HH:mm:ss";
         int attempt = 0;
         while (true) {
             try {
                 if (attempt == 0) {
                     // Use default format (MM/dd/yyyy:HH:mm:ss)
                     // Use system default timezone
-                    date = this.parseDate(time, "MM/dd/yyyy:HH:mm:ss");
+                    date = this.parseDate(time, defaultFormat);
                 }
                 else if (attempt == 1) {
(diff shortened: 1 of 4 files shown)


eemhu:
#454 has some changes for timestamp parsing, perhaps that one should be merged first?

Abigael-JT:
Yes that's right, we will revisit these changes after that PR is merged

Abigael-JT:
Rebased and updated changes after #454  was merged

src/main/java/com/teragrep/pth10/ast/DefaultTimeFormat.java
```diff
@@ -97,7 +98,10 @@ public Date parse(String time) {
             catch (ParseException ignored) {
             }
         }
-        throw new RuntimeException("TimeQualifier conversion error: <" + time + "> can't be parsed.");
+        throw new RuntimeException(
```
eemhu, line 101: could this throw `DateTimeException` or similar and also add it to the method signature as `throws DateTimeException` ?

src/main/java/com/teragrep/pth10/ast/time/InstantTimestamp.java
```diff
@@ -72,19 +75,32 @@ public Instant instant() {
...
+            final NumberFormatException cause
+    ) {
         final String unquotedValue = new UnquotedText(new TextString(value)).read(); // erase the possible outer quotes
         final Instant timevalue;
         if (timeFormatString == null || timeFormatString.isEmpty()) {
-            timevalue = new DefaultTimeFormat().parse(unquotedValue).toInstant();
+            try {
+                timevalue = new DefaultTimeFormat().parse(unquotedValue).toInstant();
+            }
+            catch (final RuntimeException ex) {
```
eemhu, line 98: (continues from previous comment) so this would catch `DateTimeException` instead of `RuntimeException` ?

src/main/java/com/teragrep/pth10/ast/DefaultTimeFormat.java
Abigael-JT: DateTimeException is thrown instead and added to method signature

eemhu (approved):
Changes look good to me, tests pass OK

elliVM:
#572 conflicts with these changes perhaps it's better to wait for a merge and apply these changes after

PR #547, opened 2025-03-18
Issue _512: Fix null return value from visitLogicalStatement

Abigael-JT:
Throw an exception if visitlogicalStatement returns a null value

### DIFF
--- a/src/main/java/com/teragrep/pth10/ast/commands/logicalstatement/LogicalStatementXML.java
+++ b/src/main/java/com/teragrep/pth10/ast/commands/logicalstatement/LogicalStatementXML.java
@@ -334,6 +334,9 @@ public class LogicalStatementXML extends DPLParserBaseVisitor<Node> {
             rv = new ElementNode(el);
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/teragrep/pth10/ast/commands/logicalstatement/LogicalStatementXML.java
```diff
@@ -334,6 +334,9 @@ else if (ctx.getChildCount() == 3) {
             rv = new ElementNode(el);
         }
 
+        else
+            throw new IllegalStateException("VisitLogicalStatement returned a Null value");
```
eemhu, line 338: Couldn't this error also appear if childCount > 3?
Also please add `{ }` for the else block

Abigael-JT: updated exception message and added {} to the else block

eemhu (approved):

PR #553, opened 2025-04-01
issue_541: Added a check and tests for empty string and null string json object

Abigael-JT:
Added a check and tests for empty string and null string properties so that fromjson() does translate and return a null


elliVM (changes requested):
Looks good and fixes the issue. 
Small change request for the  `equals()` method order.

Improvement idea:

Currently we only check if the whole JSON was just `"null"` but JSON can also have null values. We could check for those with a regex, for example something like:
`jsonString.trim().matches(".:*null(,|}|$).*");`.

eemhu (changes requested):
Please run `spotless`.
Also the helper methods could be set to `private` instead of `public` visibility.

Abigael-JT:
ran spotless and the helper methods are set to private

eemhu (approved):

PR #572, opened 2025-04-23
InstantTimestampTest test timezone fix

### LINKED ISSUE #571: InstantTimestampTest test depends on system timezone
**Describe the bug**
InstantTimestamp uses system timezone by default. This results that the test depends on the timezone where it is run. showed up in github workflow with +2 hour difference to expected values.
 
**Expected behavior**
Tests should work in any timezone
 
**How to reproduce**
Change timezone before test and run
 
**Software version**
8.4.2
 
 
**Additional context**
[Failed Run](https://github.com/teragrep/pth_10/actions/runs/14593907666/job/40935152032)

elliVM:
Fix #571 by setting the test cases expected timezone before running the test methods and return to system original default after


kortemik:
this alters the jvm state as setDefault for TimeZone will change the state globally and modifying global state in a test case is a no-go. please see https://docs.oracle.com/javase/8/docs/api/java/util/TimeZone.html#getDefault--

eemhu:
I wonder if it would be better to add an additional ctor to `InstantTimestamp` that allows defining the timezone. This ctor could be used in tests.

elliVM:
Will refactor to have a constructor with a timezone and otherwise use system default timezone

eemhu (commented):
Changes look good to me, but should the `DefaultTimeFormatTest` and `DPLTimeFormatTest` also be changed to use the new ctor?

elliVM:
yea I will update the other tests also

src/test/java/com/teragrep/pth10/DefaultTimeFormatTest.java
```diff
@@ -47,26 +47,23 @@
...
 
+import java.time.ZoneId;
 import java.util.TimeZone;
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class DefaultTimeFormatTest {
 
-    @BeforeEach
-    void enforceTimeZone() {
-        TimeZone.setDefault(TimeZone.getTimeZone("Europe/Helsinki"));
```
elliVM, line 61: removed setting default TimeZone, `Europe/Helsinki` was changing depending on daylight saving time

src/test/java/com/teragrep/pth10/DefaultTimeFormatTest.java
```diff
@@ -102,16 +99,16 @@ void ISO8601_test() {
...
-        long actual = new DefaultTimeFormat().getEpoch(time);
+        long actual = new DefaultTimeFormat(expectedTimeZone).getEpoch(time);
 
         Assertions.assertEquals(expected, actual);
     }
 
     @Test
     void defaultTimeFormatTest_2() {
         String time = "04/16/2020:10:25:40";
-        long expected = 1587021940L;
```
elliVM, line 113: test date was during daylight saving time of  `GMT+3`  so removed one hour from epoch after setting all test cases to `GMT+2`

src/test/java/com/teragrep/pth10/DefaultTimeFormatTest.java
elliVM: This was actually causing the already mentioned issue of modifying the system timezone

elliVM:
Made https://github.com/teragrep/pth_10/issues/573 for other test cases still using `TimeZone.setDefault("Europe/Helsinki")`

eemhu (approved):
builds ok, LGTM

src/main/java/com/teragrep/pth10/ast/time/InstantTimestamp.java
```diff
@@ -54,15 +54,22 @@
 import java.text.ParseException;
 import java.time.Instant;
 import java.util.Objects;
+import java.util.TimeZone;
 
 public final class InstantTimestamp implements DPLTimestamp {
 
     private final String value;
     private final String timeformat;
+    private final TimeZone timeZone;
```
kortemik, line 63: instant does not have a timezone by it's definition. Instant has a [atZone](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#atZone-java.time.ZoneId-)([ZoneId](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html) zone), please consider refactoring the method that uses the timeZone to take this ZoneId/TimeZone as it's parameter.

kortemik (changes requested):

elliVM:
Encountered another issue while refactoring.
 If the value for example `2024-10-31T10:10:10Z` contains zone information like `Z` zulu time (UTC) it will take precedence over `SimpleDateFormat.setTimezone` when calling `SimpleDateFormat.parse(time)`. Meaning that the resulting instant is always for UTC and the method `ZoneId` value is ignored. This only happens if `DefaultTimeFormat` is used meaning that format is empty.
 
I will implement it to throw an logger warning in these cases but I can elevate it to an exception if we want to

elliVM:
Some DPL tests have time value with lower case `z` at the e.g.  `earliest=2020-12-12T00:00:00z`. I suspect it was meant to set a timezone like UTC? This was ignored by the old `SimpleDateFormat` when parsing the value. In the documentation `z` was used only in the format as a "general time zone" but in a parsed value it has no function. 
Replaced the legacy `SimpleDateFormat` with a `DateTimeFormatter`, this means that the `z` input now throws an exception instead of being ignored.

Fix?
- We can change lowercase `z` to `Z` in java and this would apply a `UTC` to these values and ignore the system default.
- Change test cases to use `Z` instead of `z`.
- Enforce the zone from the class member field (System default timezone outside of test cases). This would remove the ability to set timezones in the values when no time format is given.

elliVM:
rebased

elliVM:
Refactoring time relative objects

src/main/java/com/teragrep/pth10/ast/commands/transformstatement/convert/Ctime.java
```diff
@@ -63,15 +66,14 @@ public class Ctime implements UDF2<String, String, String> {
...
-        Long e = Long.valueOf(epoch);
-
-        Date date = new Date(e * 1000L);
-        DateFormat format = new DPLTimeFormat(tf).createSimpleDateFormat();
-        format.setTimeZone(TimeZone.getTimeZone("Etc/UTC"));
-        String formatted = format.format(date);
-
-        return formatted;
+    public String call(String epoch, String timeformat) throws Exception {
+        final ZoneId utcZoneId = ZoneId.of("UTC");
```
kortemik, line 70: please create an issue about optimizing this, string conversion to time might be costly operation which could be moved to private final class member field. the same goes for the timeFormat which is unlikely to differ across the calls to the Ctime udf.

elliVM: created https://github.com/teragrep/pth_10/issues/584

elliVM:
rebased

elliVM:
Refactored relative time objects

src/main/java/com/teragrep/pth10/ast/time/InstantTimestamp.java
elliVM: Refactored timestamp objects to return `ZonedDateTime` from which further instant conversion can be done when required by other objects.

src/main/java/com/teragrep/pth10/ast/time/DefaultFormatAbsoluteTimestamp.java
```diff
@@ -43,66 +43,68 @@
...
+        LOGGER.debug("Parsing value <{}> using default formats", value);
         for (final String format : formats) {
+            LOGGER.debug("Trying format <{}>", format);
+            final AbsoluteTimestamp absoluteTimestamp = new AbsoluteTimestamp(value, format, zoneId);
             try {
-                return parseDate(time, format);
+                return absoluteTimestamp.zonedDateTime();
             }
-            catch (ParseException ignored) {
+            catch (final RuntimeException ex) {
```
eemhu, line 96: why was this changed from `ParseException` to `RuntimeException`?

src/main/java/com/teragrep/pth10/ast/time/OffsetTimestamp.java
```diff
@@ -0,0 +1,186 @@
...
+                updatedTime = startTime.plusMonths(amount);
+                break;
+            case YEARS:
+                updatedTime = startTime.plusYears(amount);
+                break;
+            default:
+                throw new RuntimeException("Unsupported unit");
+        }
+
+        // ensure that year is between 1000-9999
```
eemhu, line 115: should add here that it is due to an archiver limitation. it expects years to be 4 digits long

src/main/java/com/teragrep/pth10/ast/time/RelativeOffset.java
```diff
@@ -1,134 +0,0 @@
...
-                case YEARS:
-                    ldt = ldt.plusYears(amount);
-                    time = ldt.atZone(ZoneId.systemDefault()).toInstant();
-                    break;
-                default:
-                    throw new RuntimeException("Relative timestamp contained an invalid time unit");
-            }
-        }
-        catch (ArithmeticException ae) {
-            // on overflow, check positivity/negativity and pin to max/min
```
eemhu, line 111: is this case handled in the new object(s)?

src/main/java/com/teragrep/pth10/ast/time/RelativeTimestamp.java
```diff
@@ -45,41 +45,57 @@
...
+    public RelativeTimestamp(final ValidRelativeTimestampText offsetString, final ZonedDateTime baseTime) {
+        this.offsetString = offsetString;
+        this.baseTime = baseTime;
+    }
 
-        if (offset != null) {
-            time = offset.addOffset(time);
+    @Override
+    public ZonedDateTime zonedDateTime() {
+        if (isStub()) {
```
eemhu, line 74: I think Stub objects are not supposed to have any functionality, so perhaps rename `isStub` to `isValid` or something else.

src/main/java/com/teragrep/pth10/ast/time/SnappedTimestamp.java
```diff
@@ -0,0 +1,286 @@
...
+        SATURDAY,
+        SUNDAY,
+        MONTHS,
+        QUARTER,
+        YEARS
+    }
+
+    private final String timeStampString;
+    private final ZonedDateTime startTime;
+    private final Pattern subStringPattern = Pattern.compile("@([a-zA-Z])([+-]\\d+[a-zA-Z]+)");
```
eemhu, line 90: the regex might not accept w0-w7 units, but it might've been like this before also

src/main/java/com/teragrep/pth10/ast/time/ValidOffsetAmountText.java
```diff
@@ -0,0 +1,137 @@
...
+    private String betweenMinMax(final String input) {
+        final String betweenMinMaxString;
+        final long value;
+        try {
+            value = Long.parseLong(input);
+        }
+        catch (final NumberFormatException e) {
+            throw new RuntimeException("could not parse <" + input + "> to long. " + e.getMessage());
+        }
+        final long maxValue = 999999999L;
```
eemhu, line 122: where does this number come from?

src/main/java/com/teragrep/pth10/ast/time/ValidRelativeTimestampText.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class ValidRelativeTimestampText implements Text {
+
+    private final Text origin;
+    private final Pattern pattern;
+
+    public ValidRelativeTimestampText(final Text origin) {
+        this(origin, Pattern.compile("^(([-+])(\\d*[A-Za-z]+))?(@[A-Za-z]+([-+])?[\\dA-Za-z]*)?"));
```
eemhu, line 59: same here about the w0-w7 but i would leave it for another issue since it was here before these changes

src/main/java/com/teragrep/pth10/ast/time/ValidTrailingRelativeTimestampText.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ValidTrailingRelativeTimestampText.class);
+    private final Text origin;
+    private final Pattern validPattern = Pattern.compile(".*@([a-zA-Z]+)([+-]?\\d+[a-zA-Z]+)?$");
+
+    public ValidTrailingRelativeTimestampText(final Text origin) {
+        this.origin = origin;
+    }
+
+    public boolean isStub() {
```
eemhu, line 64: same about the stub object as before

src/test/java/com/teragrep/pth10/AbsoluteTimestampTest.java
```diff
@@ -0,0 +1,160 @@
...
+    public void testValidDateTimeFormat() {
+        final String value = "2025-05-15 14:45:00";
+        final String timeformat = "yyyy-MM-dd HH:mm:ss";
+        final AbsoluteTimestamp timestamp = new AbsoluteTimestamp(value, timeformat, utcZone);
+        final ZonedDateTime zonedDateTime = timestamp.zonedDateTime();
+        Assertions.assertEquals(2025, zonedDateTime.getYear());
+        Assertions.assertEquals(5, zonedDateTime.getMonthValue());
+        Assertions.assertEquals(15, zonedDateTime.getDayOfMonth());
+        Assertions.assertEquals(14, zonedDateTime.getHour());
+        Assertions.assertEquals(45, zonedDateTime.getMinute());
```
eemhu, line 69: might be worth also asserting the seconds, just in case.

src/test/java/com/teragrep/pth10/AbsoluteTimestampTest.java
```diff
@@ -0,0 +1,160 @@
...
+        String expectedMessage = "TimeQualifier conversion error <{Text '2025-05-15' could not be parsed: Invalid value for MonthOfYear (valid values 1 - 12): 15}>";
+        Assertions.assertEquals(expectedMessage, runtimeException.getMessage());
+    }
+
+    @Test
+    public void testUseValueZoneInformationForDateTime() {
+        final String value = "2024-05-06T15:30:00+02:00";
+        final String timeformat = "yyyy-MM-dd'T'HH:mm:ssXXX";
+        final AbsoluteTimestamp timestamp = new AbsoluteTimestamp(value, timeformat, utcZone);
+        final ZonedDateTime zonedDateTime = timestamp.zonedDateTime();
```
eemhu, line 102: assertions for time values?

src/main/java/com/teragrep/pth10/ast/time/SnappedTimestamp.java
elliVM: good catch I will look into accepting the week values in the regex

src/main/java/com/teragrep/pth10/ast/time/RelativeTimestamp.java
elliVM: Correct and I think I will rename the `isStub` to `isValid` in all classes

src/test/java/com/teragrep/pth10/EarliestLatestTest.java
```diff
@@ -472,214 +471,191 @@ public void customTimeformatTest() {
...
-        final Instant i = ts.toInstant();
-
         // Amount to add
         final int amount = 100;
         // Expected result
-        final long expected = i.plus(amount, ChronoUnit.SECONDS).getEpochSecond();
+        final long expected = startTime.plusSeconds(100).toEpochSecond();
 
-        RelativeTimeParser rtParser = new RelativeTimeParser();
         units.forEach(unit -> {
```
eemhu, line 482: could add an assertion for loops executed here

src/test/java/com/teragrep/pth10/EarliestLatestTest.java
```diff
@@ -472,214 +471,191 @@ public void customTimeformatTest() {
...
         final int amount = 100;
         // Expected result
-        final long expected = i.plus(amount, ChronoUnit.MINUTES).getEpochSecond();
+        final long expected = startTime.plusMinutes(100).toEpochSecond();
 
-        RelativeTimeParser rtParser = new RelativeTimeParser();
         units.forEach(unit -> {
-            String relativeTimestamp = "+" + amount + unit; //+100min etc.
-            Assertions.assertEquals(expected, rtParser.parse(relativeTimestamp).calculate(ts).getEpochSecond());
+            String relativeTimestamp = "+" + amount + unit; //+100d etc.
```
eemhu, line 501: same here

src/test/java/com/teragrep/pth10/EarliestLatestTest.java
```diff
@@ -472,214 +471,191 @@ public void customTimeformatTest() {
...
-        final Instant i = ts.toInstant();
-
         // Amount to add
         final int amount = 100;
         // Expected result
-        final long expected = i.plus(amount, ChronoUnit.HOURS).getEpochSecond();
+        final long expected = startTime.plusHours(100).toEpochSecond();
 
-        RelativeTimeParser rtParser = new RelativeTimeParser();
         units.forEach(unit -> {
```
eemhu, line 518: same

src/test/java/com/teragrep/pth10/OffsetTimestampTest.java
```diff
@@ -0,0 +1,177 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+public class OffsetTimestampTest {
+
+    private final ZoneId utcZone = ZoneId.of("UTC");
+    private final ZonedDateTime originTimestamp = ZonedDateTime.of(2025, 5, 15, 14, 45, 15, 790, utcZone);
+
```
eemhu, line 59: I think these tests should also assert the other parts of the timestamp to make sure they don't change

src/test/java/com/teragrep/pth10/RelativeTimestampTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+public class RelativeTimestampTest {
+
+    private final ZoneId utcZone = ZoneId.of("UTC");
+    private final ZonedDateTime originTimestamp = ZonedDateTime.of(2025, 5, 15, 14, 45, 15, 790, utcZone);
+
```
eemhu, line 59: I think these tests should also assert the other parts of the timestamp to make sure they don't change

src/test/java/com/teragrep/pth10/SnappedTimestampTest.java
```diff
@@ -0,0 +1,397 @@
...
+        Assertions.assertEquals(zonedDateTime, snappedTimestamp4.zonedDateTime());
+        Assertions.assertEquals(zonedDateTime, snappedTimestamp5.zonedDateTime());
+        Assertions.assertNotEquals(zonedDateTime, nonEqual.zonedDateTime());
+    }
+
+    @Test
+    public void testWeekDays() {
+        final List<String> weekDays = Collections
+                .unmodifiableList(Arrays.asList("@w1", "@w2", "@w3", "@w4", "@w5", "@w6"));
+        int loops = 1;
```
eemhu, line 205: I think initializing this with 0 makes more sense

src/test/java/com/teragrep/pth10/SnappedTimestampTest.java
```diff
@@ -0,0 +1,397 @@
...
+            loops++;
+        }
+        Assertions.assertEquals(7, loops);
+    }
+
+    @Test
+    public void testMonth() {
+        final SnappedTimestamp snappedTimestamp = new SnappedTimestamp("@months", originTimestamp);
+        final ZonedDateTime zonedDateTime = snappedTimestamp.zonedDateTime();
+        ;
```
eemhu, line 220: unnecessary semicolon

src/test/java/com/teragrep/pth10/relativeTimeTest.java
```diff
@@ -216,81 +238,6 @@ public void parseEndtimeuTest() {
 
     }
 
-    @Test
-    @DisabledIfSystemProperty(
-            named = "skipSparkTest",
-            matches = "true"
-    )
-    public void parseTimestampEarliestRelativeTest() {
```
eemhu, line 224: these tests have been moved to some other test class?

src/test/java/com/teragrep/pth10/AbsoluteTimestampTest.java
elliVM: added seconds and nano seconds

src/main/java/com/teragrep/pth10/ast/time/ValidTrailingRelativeTimestampText.java
elliVM: refactored to isValid

src/main/java/com/teragrep/pth10/ast/time/ValidRelativeTimestampText.java
elliVM: added regex support for weekdays

src/main/java/com/teragrep/pth10/ast/time/SnappedTimestamp.java
elliVM: updated regex to support week days with digits

src/main/java/com/teragrep/pth10/ast/time/SnappedTimestamp.java
```diff
@@ -0,0 +1,286 @@
...
+
+    @Override
+    public boolean isStub() {
+        return !timeStampString.contains("@");
+    }
+
+    private SnapUnit snapUnit() {
+        final String snapUnitSubstring;
+        final Matcher matcher = subStringPattern.matcher(timeStampString);
+        if (matcher.find()) {
```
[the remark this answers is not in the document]
elliVM: created own ValidSnapToTimeText object to handle the value extraction and throws exception is not valid

src/main/java/com/teragrep/pth10/ast/time/RelativeOffset.java
elliVM: added support to handle overflow situations

src/main/java/com/teragrep/pth10/ast/time/OffsetTimestamp.java
```diff
@@ -0,0 +1,186 @@
...
+                updatedTime = startTime.plusWeeks(amount);
+                break;
+            case MONTHS:
+                updatedTime = startTime.plusMonths(amount);
+                break;
+            case YEARS:
+                updatedTime = startTime.plusYears(amount);
+                break;
+            default:
+                throw new RuntimeException("Unsupported unit");
```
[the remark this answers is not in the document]
elliVM: changed to IllegalArgumentException

src/main/java/com/teragrep/pth10/ast/time/OffsetTimestamp.java
elliVM: changed to IllegalArgumentException

src/test/java/com/teragrep/pth10/AbsoluteTimestampTest.java
elliVM: added assertions

src/test/java/com/teragrep/pth10/EarliestLatestTest.java
elliVM: assertions for loops

src/test/java/com/teragrep/pth10/OffsetTimestampTest.java
elliVM: added assertions for unchanged value

src/test/java/com/teragrep/pth10/SnappedTimestampTest.java
elliVM: changed to 0 and assertions are for `@w0` to `@w7`

src/test/java/com/teragrep/pth10/SnappedTimestampTest.java
elliVM: removed

elliVM:
Rebased

elliVM:
Rebased again and set better logger levels. 
Invalid snap-to-time value after @ now throws an clear exception instead of being ignored. 
Supports week days `w0-w7` which was not the case before. `w8` and larger being invalid.
Added logging to easily trace the resulting time after all timestamps are applied. with resulting `ZonedDateTime` and epoch getting logged.

Tested most cases in QA

src/main/java/com/teragrep/pth10/ast/time/DefaultFormatAbsoluteTimestamp.java
```diff
@@ -43,66 +43,51 @@
...
-            }
-            catch (ParseException ignored) {
+    public ZonedDateTime zonedDateTime() {
+        for (final DPLTimestamp timestamp : timestamps) {
+            if (timestamp.isValid()) {
+                return timestamp.zonedDateTime();
             }
         }
-        throw new RuntimeException("TimeQualifier conversion error: <" + time + "> can't be parsed.");
+        throw new RuntimeException("TimeQualifier conversion error: can't be parsed using default formats.");
```
eemhu, line 86: this exception should include the user input for more clarity. now it might not be clear what the problem is if multiple timeQualifiers (earliest / latest) are used.

src/main/java/com/teragrep/pth10/ast/time/OffsetTimestamp.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    private long offsetAmount() {
+        final String validAmountString = new ValidOffsetAmountText(offsetText).read();
+        return Long.parseLong(validAmountString);
+    }
+
+    private OffsetUnit offsetUnit() {
+        final String validUnitString = new ValidOffsetUnitText(offsetText).read();
+        switch (validUnitString.toLowerCase()) {
```
eemhu, line 174: this might be preferred to use single point of return

src/main/java/com/teragrep/pth10/ast/time/ValidOffsetUnitText.java
```diff
@@ -0,0 +1,86 @@
...
+        final String updatedString;
+        if (originString.toLowerCase().startsWith("now")) {
+            updatedString = "now";
+        }
+        else if (originString.toLowerCase().contains("now")) {
+            throw new RuntimeException("timestamp 'now' should not have any values before it");
+        }
+        else {
+            if (!matcher.find()) {
+                throw new RuntimeException("Text did not contain a valid offset unit");
```
eemhu, line 80: perhaps also log the input

src/main/java/com/teragrep/pth10/ast/time/ValidOffsetAmountText.java
```diff
@@ -0,0 +1,134 @@
...
+        if (amountString.isEmpty() || "now".equalsIgnoreCase(amountString)) {
+            updatedString = "0";
+        }
+        else if ("+".equals(amountString)) {
+            updatedString = "1";
+        }
+        else if ("-".equals(amountString)) {
+            updatedString = "-1";
+        }
+        else if (amountString.matches(".*\\d.*")) {
```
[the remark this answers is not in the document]
elliVM: It checks if there is a single digit in the string anywhere so `-1` and `1` are both true.
I could replace it with `else if (amountString.chars().anyMatch(Character::isDigit)) {`. what do you think  would that be better?

src/main/java/com/teragrep/pth10/ast/time/ValidOffsetAmountText.java
```diff
@@ -0,0 +1,134 @@
...
+            value = Long.parseLong(input);
+        }
+        catch (final NumberFormatException e) {
+            throw new RuntimeException("could not parse <" + input + "> to long. " + e.getMessage());
+        }
+        // epoch seconds range supported by ZonedDateTime
+        final long maxEpochSeconds = 999999999L;
+        final long minEpochSeconds = -999999999L;
+        if (value > maxEpochSeconds) {
+            LOGGER.info("value over max value");
```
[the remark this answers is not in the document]
elliVM: removed these logger messages

eemhu:
> Added `equals()` and `hashCode` methods and EqualsVerifier tests. `ValidTrailingRelativeTimestampTextTest` could not be tested as EqualsVerifier was stuck in a recursive loop. Only solution I found was to make the class member field `Pattern` class `static` which I didn't want to do. Added `@Disable` tag for now with a message and comments about the issue. I can also add an issue about it after merge.

This was due to having the `Pattern` class imported from a wrong package.

elliVM:
rebased and ran spotless for updated licenses

elliVM:
after package rename rebase this version had some ClassDefNotFound issues in QA. will investigate

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/Mvrange.java
```diff
@@ -102,11 +103,11 @@ else if (stepObj instanceof String) {
             long time = start;
             rv.add(String.valueOf(time));
 
-            RelativeTimeParser rtParser = new RelativeTimeParser();
-            RelativeTimestamp rtTimestamp = rtParser.parse("+" + stepStr);
             // Go until incremented past end
             while (time < end) {
-                time = rtTimestamp.calculate(new Timestamp(time * 1000L)).getEpochSecond();
+                final ZonedDateTime startTime = ZonedDateTime.ofInstant(Instant.ofEpochSecond(time), ZoneId.of("UTC"));
```
eemhu, line 108: I think the `ZoneId.of` call should be moved outside of the loop

src/main/java/com/teragrep/pth_10/ast/time/RelativeTimestamp.java
```diff
@@ -45,41 +45,78 @@
...
+            updatedTime = offsetTimestamp.zonedDateTime();
         }
+        return updatedTime;
+    }
 
-        if (offset != null) {
-            time = offset.addOffset(time);
+    @Override
+    public boolean isValid() {
+        boolean isStub = true;
```
eemhu, line 93: change variable name to isValid

src/main/java/com/teragrep/pth_10/ast/time/ValidOffsetAmountText.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class ValidOffsetAmountText implements Text {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ValidOffsetAmountText.class);
+
+    private final Text origin;
+    private final Pattern pattern = Pattern.compile("^([+-]?\\d+)");
```
eemhu, line 61: this could be in ctor, and main ctor could be private so the pattern cannot be changed

src/test/java/com/teragrep/pth_10/DPLTimeFormatTextTest.java
```diff
@@ -0,0 +1,58 @@
...
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_10;
+
+import com.teragrep.pth_10.ast.DPLTimeFormatText;
+import nl.jqno.equalsverifier.EqualsVerifier;
+import org.junit.jupiter.api.Test;
+
+public class DPLTimeFormatTextTest {
+
```
eemhu, line 53: should this test the read function?

elliVM: added test for reads

src/main/java/com/teragrep/pth_10/ast/time/ValidOffsetAmountText.java
```diff
@@ -0,0 +1,153 @@
...
+        else if (amountString.chars().anyMatch(Character::isDigit)) {
+            final Matcher matcher = pattern.matcher(amountString);
+            if (matcher.find()) {
+                updatedString = matcher.group();
+            }
+            else {
+                throw new RuntimeException("Matcher could not find a valid offset amount from <" + amountString + ">");
+            }
+        }
+        else {
```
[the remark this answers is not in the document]
elliVM: simplified to remove nesting

src/main/java/com/teragrep/pth_10/ast/time/ValidOffsetAmountText.java
elliVM: added constructor

src/main/java/com/teragrep/pth_10/ast/time/OffsetTimestamp.java
```diff
@@ -0,0 +1,249 @@
...
+                    }
+                    break;
+                default:
+                    throw new IllegalArgumentException("Unsupported unit <" + unit + ">");
+            }
+        }
+        catch (final DateTimeException | ArithmeticException exception) {
+            LOGGER.info("Overflow exception when calculating offset <{}>", exception.getMessage());
+            if (amount > 0) {
+                LOGGER.info("Positive amount overflow, clamping to min accepted value");
```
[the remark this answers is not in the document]
elliVM: fixed comment

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/Mvrange.java
elliVM: moved utc `ZoneId` outside of the loop

src/main/java/com/teragrep/pth_10/ast/time/ValidOffsetAmountText.java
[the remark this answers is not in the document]
eemhu: what about the case where offsetAmount=+ (example `+h`)? Also, is it necessary that the `amountValueContainsDigits` is checked, wouldn't the regex catch that?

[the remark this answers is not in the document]
elliVM: refactored so digits is checked only by the regex and the boolean is removed, also replaced matcher `find()` with `lookingAt()` to ensure that matching starts at the start of the string

eemhu (changes requested):
rebase, then should be ok

eemhu (dismissed):

src/main/java/com/teragrep/pth_10/ast/time/DPLTimestampImpl.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public DPLTimestampImpl(final String value, final String timeformat) {
+        this(value, timeformat, TimeZone.getDefault().toZoneId());
+    }
+
+    public DPLTimestampImpl(final String value, final String timeformat, final ZoneId zoneId) {
+        this(new AbsoluteTimestamp(value, timeformat, zoneId), new RelativeTimestamp(value, zoneId));
+    }
+
+    public DPLTimestampImpl(final AbsoluteTimestamp absoluteTimestamp, final RelativeTimestamp relativeTimestamp) {
```
kortemik, line 78: Please describe the meaning of this ctor

src/main/java/com/teragrep/pth10/ast/commands/transformstatement/convert/Mktime.java
```diff
@@ -60,10 +60,8 @@ public class Mktime implements UDF2<String, String, String> {
 
     @Override
     public String call(String hrt, String tf) throws Exception {
-        DPLTimeFormat format = new DPLTimeFormat(tf);
-        Long rv = format.instantOf(hrt).getEpochSecond();
-
-        return rv.toString();
+        final DPLTimestampImpl timestamp = new DPLTimestampImpl(hrt, tf);
```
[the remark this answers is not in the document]
elliVM: refactored to full names

src/main/java/com/teragrep/pth_10/ast/time/DPLTimestampImpl.java
elliVM: this is the primary ctor to simply assign the encapsulated objects inside, however leaving it `public` could expose it to miss matching timestamp values so I refactored it to be a private ctor.

kortemik:
I think following design might make sense:

Timestamps are encapsulated into DPLTimestampString object.

The DPLTimestampString object takes DPLTimeFormats via ctor as a List.

DPLTimeFormat is an interface that provides DPLTimestamp from(String dplTimestampString)

DPLTimestampString has a method asDPLTimestamp which iterates over the provided DPLTimestamp.from() methods and collects a list of possible interpretations. If one and only one is found it returns it. If none is found it returns a stub. If multiple are found it returns IllegalStateException because multiple interpretations must not exist.

DPLTimeFormat implementation for DPLRelativeFormat takes a ctor of (String type, ZonedDateTime referencePoint) where the referencePoint is from the query start time. This must be provided within the DPLContext (for now) when the context is created. Other implementations of the DPLTimeFormat of course exist, such as the plain static formats.

src/main/java/com/teragrep/pth_10/ast/DPLParserCatalystVisitor.java
```diff
@@ -266,7 +267,8 @@ public Node visitRoot(DPLParser.RootContext ctx) {
         LOGGER.info("CatalystVisitor Root incoming: <{}>", ctx.getText());
         // Set DPL query into CatalystContext: used if original query string is needed for something
         catCtx.setDplQuery(ctx.getText());
-
+        // set star time of query execution
+        catCtx.setStartTime(ZonedDateTime.now());
```
elliVM, line 271: not 100% if this is the correct entry point to set the query start time, also should the timezone be parameterized? currently uses system default.

eemhu: would also remove the need for a setter

elliVM: Sounds good I'll add it to the ctor

src/main/java/com/teragrep/pth_10/ast/time/DPLTimestampString.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.stream.Collectors;
+
+public final class DPLTimestampString {
+
+    private final Logger LOGGER = LoggerFactory.getLogger(DPLTimestampString.class);
+
+    private final String timestampString;
+    private final ZonedDateTime baseTime;
+    private final List<DPLTimeFormat> defaultFormats;
+    private final List<DPLTimeFormat> userDefinedFormats;
```
eemhu, line 74: I think these do not need to be separate, because the parsing should fail if the user defined timeformat is not valid.

elliVM: Issue was that user can provide a format where a value can match with default format and the user defined format. in that chase should we just throw an exception that they clashed?

src/main/java/com/teragrep/pth_10/executor/DPLExecutorImpl.java
```diff
@@ -65,21 +65,24 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger(DPLExecutorImpl.class);
     private final Config config;
+    private final ZonedDateTime startTime;
     private final BatchCollect batchCollect;
 
     // Active query
     StreamingQuery streamingQuery = null;
 
-    public DPLExecutorImpl(Config config) {
+    public DPLExecutorImpl(Config config, ZonedDateTime startTime) {
```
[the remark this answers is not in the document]
eemhu: I don't think it is possible without changes to the pth_15 component. This implementation class is loaded via `DPLExecutorFactory` that expects one `Config` type arg ctor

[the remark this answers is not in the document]
elliVM: Then the only option currently would be to use `ZonedDateTime.now()` and system default timezone?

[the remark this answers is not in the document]
eemhu: I think the most ideal option would be to set it in `DPLExecutorImpl.interpret()` method.

src/main/java/com/teragrep/pth_10/ast/time/OffsetTimestamp.java
```diff
@@ -0,0 +1,254 @@
...
+        return false;
+    }
+
+    private long offsetAmount() {
+        final String validAmountString = new ValidOffsetAmountText(offsetText).read();
+        return Long.parseLong(validAmountString);
+    }
+
+    private OffsetUnit offsetUnit() {
+        final String validUnitString = new ValidOffsetUnitText(offsetText).read();
```
eemhu, line 179: maybe these two should be their own objects?

src/test/java/com/teragrep/pth_10/DPLTimestampStringTest.java
```diff
@@ -0,0 +1,224 @@
...
+        Assertions.assertEquals(7, zonedDateTime.getDayOfMonth());
+        Assertions.assertEquals(13, zonedDateTime.getHour());
+        Assertions.assertEquals(20, zonedDateTime.getMinute());
+        Assertions.assertEquals(0, zonedDateTime.getSecond());
+        Assertions.assertEquals(0, zonedDateTime.getNano());
+        Assertions.assertEquals(utcZone, zonedDateTime.getZone());
+    }
+
+    @Test
+    public void testUserDefinedFormat() {
```
eemhu, line 190: test invalid user defined format also

src/main/java/com/teragrep/pth_10/executor/DPLExecutorImpl.java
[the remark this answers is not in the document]
elliVM: updated pth_15 version and added start time to the method

elliVM:
Rebased

PR #630, opened 2025-07-16
Improved error message for invalid  spath input

### LINKED ISSUE #137: Create better error message spath numeric
ID <533>
Reporter <@JohnJacobGuy>

index=example | spath

 results in:

java.lang.RuntimeException: org.apache.spark.sql.streaming.StreamingQueryException: Job aborted due to stage failure: ....

'example' index contains only numeric data (like 12.34) but should throw more appropriate error.

#403

Abigael-JT:
## Description

- Added a valid Json Object check and improved error message for invalid spath command input
- Checks jsonElement in Spath to not be null
- Added tests

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality


elliVM:
The spath `call` method currently catches all exceptions that it throws. This will result in a silent failure which will then obscure the real reason as stated in the issue.
Would it be better if we don't catch all exceptions and fail fast in those cases where the input is not acceptable.

elliVM (dismissed):
LGTM

eemhu (changes requested):
I think instead of an error, it should return empty result
See the exception handling as an example.

https://github.com/teragrep/pth_10/pull/630/files#diff-16e63a38b9ccb7ccc0ddd407b3a83a3f1dd1e63cb4bc8a2771b24941c78010c7R176-R187

Abigael-JT:
combined with changes from PR #621

eemhu:
> @eemhu wouldn't that result be confusing though? it's probably helpful to know why the result is empty or not as expected

It would be similar behavior to the other cases, where JSON parsing fails. And also if there is one row with invalid data it will stop the whole query even if the other rows would parse successfully.

Abigael-JT:
> I think instead of an error, it should return empty result See the exception handling as an example.
> 
> https://github.com/teragrep/pth_10/pull/630/files#diff-16e63a38b9ccb7ccc0ddd407b3a83a3f1dd1e63cb4bc8a2771b24941c78010c7R176-R187

updated to return an empty result if the input is not a  JsonObject

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/Spath.java
```diff
@@ -112,6 +112,18 @@ public Map<String, String> call(String input, String spathExpr, String nameOfInp
             final Gson gson = new Gson();
             final JsonElement jsonElem = gson.fromJson(input, JsonElement.class);
 
+            if (jsonElem == null || jsonElem.isJsonNull()) {
+                LOGGER.warn("Json input was 'null', returning empty result");
+                return result;
+            }
+
```
eemhu, line 119: these two if-blocks claim to return empty result, however they are different:
first one returns an empty map
second one returns a map with one key-value pair, where the value is null

eemhu: see line 184 what kind of result is expected when an error occurs

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/Spath.java
```diff
@@ -112,6 +112,24 @@ public Map<String, String> call(String input, String spathExpr, String nameOfInp
...
+            if (!jsonElem.isJsonObject()) {
```
eemhu: JsonArray is valid also

`"[{\"a\":\"b\", \"b\":\"c\"}, {\"a2\":\"b2\"}]"`

should auto-extract all the values, with the keys being the column names.

```
[].a -> b
[].b -> c
[].a2 -> b2
```

Abigael-JT: added support for extracting values from JsonArray

Abigael-JT:
Added new objects for parsing Json and Xml separately

creating tests for the new objects

Abigael-JT:
created issue #857  for the disabled test **testSpathXmlInputDuplicatekeysWithPath()** in SpathXml.Java

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/SpathXml.java
```diff
@@ -0,0 +1,238 @@
...
+     * @param inputColumn     name of input column
+     * @param outputColumn    name of output column
+     */
+    public SpathXml(
+            final String input,
+            final String spathExpression,
+            final String inputColumn,
+            final String outputColumn
+    ) {
+        this(input, spathExpression, inputColumn, outputColumn, new NullValue());
```
eemhu, line 95: NullValue is a object that can be configured and only the one provided in DPLParserCatalystContext should be used so that the set configuration is not lost.

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/SpathFlatJson.java
```diff
@@ -0,0 +1,149 @@
...
+        final String value = new UnquotedText(new TextString(StringEscapeUtils.unescapeJson(primitive.toString())))
+                .read();
+        final String key = new QuotedText(new TextString(path), "`").read();
+        result.put(key, value);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        final boolean isEquals;
+        if (this == o)
```
eemhu: if-else blocks should always use curly braces { }

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/SpathXml.java
```diff
@@ -0,0 +1,238 @@
...
+                // get parent of parent
+                parent = parent.getParentNode();
+                isFirst = false;
+            }
+
+            // if there are multiple columns of the same name, add value to existing column
+            final String quotedColName = new QuotedText(new TextString(colName), "`").read();
+            if (map.containsKey(quotedColName)) {
+                map
+                        .computeIfPresent(quotedColName, (k, existingValue) -> existingValue.concat("\n").concat(rootNode.getTextContent()));
```
eemhu, line 203: This probably should provide an array instead of line break separated list. (To be properly functional as a multivalue) Perhaps for another issue since it probably has been this way before.

src/test/java/com/teragrep/pth_10/SpathFlatJsonTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    @Test
+    void testSpathInvalidNumericJsonInput() {
+        final String input = "12.34";
+        final Map<String, String> result = new SpathFlatJson(input).asMap();
+        Assertions.assertTrue(result.isEmpty());
+    }
+
+    @Test
+    void testSpathNullSpathExpression() {
```
eemhu: where is the null spath expression?

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/SpathXml.java
Abigael-JT: created issue https://github.com/teragrep/pth_10/issues/879

src/test/java/com/teragrep/pth_10/SpathFlatJsonTest.java
Abigael-JT: Wrong name, updated

PR #635, opened 2025-07-18
Improved unit tests assertions

Abigael-JT:
## Description
Improved assertions for the tests that needs to use loops and iteration.
Removed unnecessary braching

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/src/test/java/com/teragrep/pth_10/commandTest.java
+++ b/src/test/java/com/teragrep/pth_10/commandTest.java
@@ -204,6 +204,7 @@ public class commandTest {
         this.streamingTestUtil.performDPLTest(q, this.testFile, res -> {
             List<Row> sourcetypeCol = res.select("sourcetype").distinct().collectAsList();
             Assertions.assertEquals(9, res.count());
+            Assertions.assertEquals(1, sourcetypeCol.size());
             Assertions.assertEquals("teragrep version", sourcetypeCol.get(0).get(0).toString());
 
             List<Row> rawCol = res.select("_raw").collectAsList();


src/test/java/com/teragrep/pth10/MakeresultsTransformationTest.java
```diff
@@ -121,9 +121,13 @@ public void makeresults_Annotate_Test() {
             // assert all of them to be null
             rows.forEach(row -> {
                 Assertions.assertEquals(6, row.length());
+
+                int loopCount = 0;
```
elliVM, line 125: Here we have already asserted that `row.length()` is `6` and we use that as the limit in our for loop, in this case the loop count check is redundant

src/test/java/com/teragrep/pth10/RangemapTransformationTest.java
```diff
@@ -203,11 +203,12 @@ public void testrangemapMultiValue() {
                             Assertions.assertEquals(2, resultList.size());
                             List<String> expected = Arrays.asList("lo", "hi");
 
+                            int expectedLoops = 0;
                             for (String res : resultList) {
```
elliVM, line 207: here we could also just test the equality of the `expected` and `resultList`, just ensure the expected has the correct order. Then we could remove the for loop completely

src/test/java/com/teragrep/pth10/RegexTransformationTest.java
```diff
@@ -117,8 +120,9 @@ public void testRegexFieldNotEqual() {
     )
     public void testRegexFieldEqual() {
         streamingTestUtil.performDPLTest("index=index_A | regex _raw = \"data data\"", testFile, ds -> {
-            int size = ds.collectAsList().size();
-            Assertions.assertTrue(size > 1);
+            List<Row> resList = ds.select("_raw").distinct().collectAsList();
```
elliVM, line 123: Here we could clear up the assertion statement by mapping the `resList` to String. Also we can directly test equality now and not the  `contains()` method

```
List<String> resList = ds.select("_raw").distinct().collectAsList()
                    .stream()
                    .map(r -> r.get(0))
                    .map(Object::toString)
                    .collect(Collectors.toList());
            Assertions.assertEquals(1, resList.size());
            Assertions.assertEquals("data data", resList.get(0));
```

same can be done in some other tests in this class

src/test/java/com/teragrep/pth10/chartTransformationTest.java
```diff
@@ -122,17 +122,23 @@ void testChartCountWithAsAndSplitBy() {
...
 
+            List<String> resultList = res
+                    .collectAsList()
+                    .stream()
+                    .map(r -> r.mkString(","))
+                    .collect(Collectors.toList());
+
             // 3 first rows are earlier than where _index_earliest is set to
             List<String> expectedValues = new ArrayList<>();
+            int loopCount = 0;
```
elliVM, line 133: The loop check is redundant for fixed range for loops. I think the  `expectedValues.size()` assertion is enough here.

src/test/java/com/teragrep/pth10/commandTest.java
```diff
@@ -174,18 +174,16 @@ void testTeragrep() {
 
         this.streamingTestUtil.performDPLTest(q, this.testFile, res -> {
             List<Row> sourcetypeCol = res.select("sourcetype").collectAsList();
-            for (Row r : sourcetypeCol) {
-                Assertions.assertTrue(r.getString(0).contains("teragrep version"));
-            }
+            Assertions.assertTrue(sourcetypeCol.get(0).get(0).toString().contains("teragrep version"));
```
elliVM, line 177: this could be `assertEquals` instead of using the `contains()`

elliVM (commented):
Good changes, I left some comments where I found improvement ideas or corrections.

src/test/java/com/teragrep/pth_10/commandTest.java
```diff
@@ -174,18 +174,16 @@ void testTeragrep() {
 
         this.streamingTestUtil.performDPLTest(q, this.testFile, res -> {
             List<Row> sourcetypeCol = res.select("sourcetype").collectAsList();
-            for (Row r : sourcetypeCol) {
-                Assertions.assertTrue(r.getString(0).contains("teragrep version"));
-            }
+            Assertions.assertEquals("teragrep version", sourcetypeCol.get(0).get(0).toString());
```
eemhu, line 177: these no longer assert all rows?

src/test/java/com/teragrep/pth_10/translationTests/SortTest.java
```diff
@@ -156,9 +156,11 @@ void testSortTranslation2() {
         expected.add(sbc3);
         expected.add(sbc4);
 
+        int expectedLoops = 0;
```
eemhu, line 159: expected -> executed

src/test/java/com/teragrep/pth_10/commandTest.java
Abigael-JT: results has  just one row

Abigael-JT: done

src/test/java/com/teragrep/pth_10/RangemapTransformationTest.java
```diff
@@ -201,13 +201,9 @@ public void testrangemapMultiValue() {
                             Assertions.assertEquals(1, result.size());
                             List<String> resultList = result.get(0).getList(0);
                             Assertions.assertEquals(2, resultList.size());
-                            List<String> expected = Arrays.asList("lo", "hi");
+                            List<String> expected = Arrays.asList("hi", "lo");
```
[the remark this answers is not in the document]
Abigael-JT: good point,  and probably not a guarantee,  makes sense to assert that the list contains the expected strings instead

src/test/java/com/teragrep/pth_10/JoinTransformationTest.java
```diff
@@ -312,17 +312,17 @@ public void join0MaxOverwriteExplicitTest() {
...
 
-                            int notNulls = 0;
-                            for (Row r : listOfAColumn) {
-                                String val = r.getString(0);
-                                Assertions.assertNotNull(val, "All rows should have a valid value (non-null) !");
-                                notNulls++;
-                            }
-                            Assertions.assertEquals(10, notNulls, "Should execute 10 loops");
+                            Assertions.assertEquals(10, listOfAColumn.size(), "Should have 10 rows");
+                            Assertions.assertNotNull(listOfAColumn, "All rows should have a valid value (non-null) !");
```
eemhu, line 325: line 325 not necessary check

PR #693, opened 2025-10-31
Issue 647: Bloom create and update now returns only status messages.

eemhu:
## Description

teragrep exec bloom create
teragrep exec bloom update
will now return "Bloom filter operation CREATE was completed." or "Bloom filter operation UPDATE was completed." with _time, instead of the whole dataset to UI.

same could be done for hdfs save, but perhaps that one should come in a separate pull request. -> #699 

#### General

- [ ] I have checked that my test files and functions have meaningful names.
- [ ] I have checked that each test tests only a single behavior.
- [ ] I have done happy tests.
- [ ] I have tested only my own code.
- [ ] I have tested at least all public methods. 

#### Assertions

- [ ] I have checked that my tests use assertions and not runtime overhead.
- [ ] I have checked that my tests end in assertions.
- [ ] I have checked that there is no comparison statements in assertions.
- [ ] I have checked that assertions are in tests and not in helper functions.
- [ ] I have checked that assertions for iterables are outside of for loops and both sides of the iteration blocks.
- [ ] I have checked that assertions are not tested inside consumers.

#### Testing Data

#### Statements

- [ ] I have checked that my tests do not use throws for exceptions.
- [ ] I have checked that my tests do not use try-catch statements.
- [ ] I have checked that my tests do not use if-else statements.

#### Java

- [ ] I have checked that my tests for Java uses JUnit library.
- [ ] I have checked that my tests for Java uses JUnit utilities for parameters.

#### Other

### Code Quality

### DIFF
--- a/src/main/java/com/teragrep/pth_10/datasources/CustomDatasetImpl.java
+++ b/src/main/java/com/teragrep/pth_10/datasources/CustomDatasetImpl.java
@@ -50,10 +50,6 @@ import org.apache.spark.sql.*;
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder;
 import org.apache.spark.sql.catalyst.encoders.RowEncoder;
 import org.apache.spark.sql.execution.streaming.MemoryStream;
-import org.apache.spark.sql.streaming.DataStreamWriter;
-import org.apache.spark.sql.streaming.OutputMode;
-import org.apache.spark.sql.streaming.StreamingQuery;
-import org.apache.spark.sql.streaming.StreamingQueryException;
 import org.apache.spark.sql.types.StructType;
 import scala.Option;
 
@@ -82,22 +78,14 @@ public final class CustomDatasetImpl implements CustomDataset {
     }
 
     @Override
-    public Dataset<Row> asStreamingDataset() throws StreamingQueryException {
+    public Dataset<Row> asStreamingDataset() {
         final SQLContext sqlContext = catCtx.getSparkSession().sqlContext();
         final ExpressionEncoder<Row> encoder = RowEncoder.apply(schema);
         final MemoryStream<Row> rowMemoryStream = new MemoryStream<>(1, sqlContext, Option.apply(1), encoder);
 
-        final Dataset<Row> rowDataset = rowMemoryStream.toDF();
-        final String queryName = "custom_dataset_" + ((int) (Math.random() * 100000));
-
-        final DataStreamWriter<Row> writer = rowDataset.writeStream().format("memory").outputMode(OutputMode.Append());
-
         rowMemoryStream.addData(new Rows(values).asSeq());
 
-        final StreamingQuery sq = this.catCtx.getInternalStreamingQueryListener().registerQuery(queryName, writer);
-        sq.awaitTermination();
-
-        return rowDataset;
+        return rowMemoryStream.toDF();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


elliVM:
New objects look solid - I was thinking about adding a `CustomDatasetStep` object to apply the `CustomDataset` at a step level without having to write it into the logic of the `TeragrepBloomStep`.
It could be simply wrap the custom dataset class as a step `new CustomDatasetStep(new CustomDataset);`

So the `CustomDatasetStep` would be added in the `visitT_bloomOptionParameter()` method of the `TeragrepTransoformation` for update and create commands.
NOTE: this should be added only for the bloom update and create commands but they already have their own step logic in place. 
Result example would be `new StepListNode(Arrays.asList(aggregateStep, bloomStepWithRegexAndTable, customDatasetStep));` the same could be then done easily for other steps that require just a simple message after finishing.

Let me know if you think it's an good idea.

eemhu:
> New objects look solid - I was thinking about adding a `CustomDatasetStep` object to apply the `CustomDataset` at a step level without having to write it into the logic of the `TeragrepBloomStep`. It could be simply wrap the custom dataset class as a step `new CustomDatasetStep(new CustomDataset);`
> 
> So the `CustomDatasetStep` would be added in the `visitT_bloomOptionParameter()` method of the `TeragrepTransoformation` for update and create commands. NOTE: this should be added only for the bloom update and create commands but they already have their own step logic in place. Result example would be `new StepListNode(Arrays.asList(aggregateStep, bloomStepWithRegexAndTable, customDatasetStep));` the same could be then done easily for other steps that require just a simple message after finishing.
> 
> Let me know if you think it's an good idea.

Sounds good, I will try to implement

src/main/java/com/teragrep/pth_10/datasources/CustomDatasetImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+        final Dataset<Row> rowDataset = rowMemoryStream.toDF();
+        final String queryName = "custom_dataset_" + ((int) (Math.random() * 100000));
+
+        final DataStreamWriter<Row> writer = rowDataset.writeStream().format("memory").outputMode(OutputMode.Append());
+
+        rowMemoryStream.addData(new Rows(values).asSeq());
+
+        final StreamingQuery sq = this.catCtx.getInternalStreamingQueryListener().registerQuery(queryName, writer);
+        sq.awaitTermination();
```
kortemik: this will block processing so is it intended?

src/main/java/com/teragrep/pth_10/datasources/CustomDatasetImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public Dataset<Row> asStreamingDataset() throws StreamingQueryException {
+        final SQLContext sqlContext = catCtx.getSparkSession().sqlContext();
+        final ExpressionEncoder<Row> encoder = RowEncoder.apply(schema);
+        final MemoryStream<Row> rowMemoryStream = new MemoryStream<>(1, sqlContext, Option.apply(1), encoder);
+
+        final Dataset<Row> rowDataset = rowMemoryStream.toDF();
+        final String queryName = "custom_dataset_" + ((int) (Math.random() * 100000));
```
kortemik: is uuid possible?

PR #715, opened 2025-11-21
Spark 3.5.7 Update & Necessary Changes

eemhu:
## Description

* Updates Spark 3.4.0 to 3.5.7
* Makes necessary changes to DPLInternalStreamingQueryListener to successfully stop idle queries.
* Disables TeragrepKafkaTest as the testing library is not compatible with Kafka 3.5 and above.
* Excludes dependencies with conflicting version numbers.
* Changed `RowEncoder.apply` usages to `ExpressionEncoder.apply` as it is no longer available in Spark 3.5 and above.
* pth_06 PR https://github.com/teragrep/pth_06/pull/250 should be merged and released first, then the version can be updated here and this PR can be merged and released. Note that ZEP-01/PTH-07 component will need some changes to the query stopping functionality as seen in this PR in the DPLInternalStreamingQueryListener. --> The listener used is included in this project, fixed also in this PR.

#### General

- [ ] I have checked that my test files and functions have meaningful names.
- [ ] I have checked that each test tests only a single behavior.
- [ ] I have done happy tests.
- [ ] I have tested only my own code.
- [ ] I have tested at least all public methods. 

#### Assertions

- [ ] I have checked that my tests use assertions and not runtime overhead.
- [ ] I have checked that my tests end in assertions.
- [ ] I have checked that there is no comparison statements in assertions.
- [ ] I have checked that assertions are in tests and not in helper functions.
- [ ] I have checked that assertions for iterables are outside of for loops and both sides of the iteration blocks.
- [ ] I have checked that assertions are not tested inside consumers.

#### Testing Data

#### Statements

- [ ] I have checked that my tests do not use throws for exceptions.
- [ ] I have checked that my tests do not use try-catch statements.
- [ ] I have checked that my tests do not use if-else statements.

#### Java

- [ ] I have checked that my tests for Java uses JUnit library.
- [ ] I have checked that my tests for Java uses JUnit utilities for parameters.

#### Other

### Code Quality

### DIFF
--- a/src/main/java/com/teragrep/pth_10/ast/DPLInternalStreamingQueryListener.java
+++ b/src/main/java/com/teragrep/pth_10/ast/DPLInternalStreamingQueryListener.java
@@ -54,9 +54,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeoutException;
 
 /**
@@ -70,7 +70,7 @@ public class DPLInternalStreamingQueryListener extends StreamingQueryListener im
     /**
      * Map containing all the queryName-DPLInternalStreamingQuery k,v pairs
      */
-    private final Map<UUID, DPLInternalStreamingQuery> _queries = new HashMap<>();
+    private final Map<UUID, DPLInternalStreamingQuery> _queries = new ConcurrentHashMap<>();
 
     public DPLInternalStreamingQueryListener() {
         super();


kortemik:
@eemhu i released pth_06 4.0.0. with the changes

eemhu:
Updated to pth_06 4.0.0 and added necessary changes for the DPLStreamingQueryListener. Awaiting for Spark 3.5 enabled QA for further testing.

kortemik (changes requested):

kortemik (approved):

eemhu:
Appears that queries containing indexStatements are stopped correctly,
however queries that use data generating commands are not.

src/main/java/com/teragrep/pth_10/executor/SourceStatus.java
```diff
@@ -64,7 +64,10 @@ public boolean isQueryDone(StreamingQuery outQ) {
             boolean queryDone = true;
             for (int i = 0; i < outQ.lastProgress().sources().length; i++) {
                 final String description = outQ.lastProgress().sources()[i].description();
-                if (!description.startsWith(com.teragrep.pth_06.ArchiveMicroStreamReader.class.getName().concat("@"))) {
+                if (
+                    !description.startsWith(com.teragrep.pth_06.ArchiveMicroStreamReader.class.getName().concat("@"))
```
eemhu: would checking this description be necessary if we would just check the kafka continuousProcessing flag ?

eemhu: However testing locally without InputPartitions, the queryIdleEvent was only triggered at the end of the query when the last offset was encountered.

PR #736, opened 2025-12-08
Adds wildcard support for fields command

### LINKED ISSUE #737: Fields command does not work correctly with wildcard
**Describe the bug**

%dpl
index=abc earliest=-5y
| fields s*

Returns this error message:
`[PARSE_SYNTAX_ERROR] Syntax error at or near end of input.(line 1, pos 2)`

`== SQL ==`
`s*`
`--^^^`

Also the query below returns all columns, it does not filter out columns starting with underscore "_" as expected

%dpl
index=abc earliest=-5y
| fields - _*

**Expected behavior**

It should return the columns that matches the specified pattern
 
**How to reproduce**

Run the above queries
 
**Screenshots**

 
**Software version**

pth_10 version: 10.3.4
pth_03 version: 9.3.2
 
**Desktop (please complete the following information if relevant):**
 - OS: 
 - Browser: 
 - Version: 
 
**Additional context**

Abigael-JT:
- Added new object FilteredColumns.java to check for wildcards in the given fields
- Added wildcard tests for fields command
- used FilteredColumns.java object to check for wildcards in table commands' fields

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/src/main/java/com/teragrep/pth_10/ast/FilteredColumns.java
+++ b/src/main/java/com/teragrep/pth_10/ast/FilteredColumns.java
@@ -45,16 +45,18 @@
  */
 package com.teragrep.pth_10.ast;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import static org.apache.log4j.builders.appender.SocketAppenderBuilder.LOGGER;
+public final class FilteredColumns {
 
-public class FilteredColumns {
-
+    private static final Logger LOGGER = LoggerFactory.getLogger(FilteredColumns.class);
     private final String[] columns;
     private final String wc;
 
@@ -65,7 +67,7 @@ public class FilteredColumns {
      * @param columns array of column names
      */
 
-    public FilteredColumns(String wc, String[] columns) {
+    public FilteredColumns(final String wc, final String[] columns) {
         this.columns = columns;
         this.wc = wc;
     }
@@ -79,7 +81,7 @@ public class FilteredColumns {
         final StringBuilder regexBuilder = new StringBuilder();
         final String regex;
 
-        for (char c : wc.toCharArray()) {
+        for (final char c : wc.toCharArray()) {
             if (c == '*') {
                 // On wildcard, get preceding content and quote it
                 // Also clear quotablePartBuilder and add regex any char wildcard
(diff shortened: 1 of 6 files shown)


src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/FieldsTransformation.java
```diff
@@ -78,18 +78,23 @@ public Node visitFieldsTransformation(DPLParser.FieldsTransformationContext ctx)
 
     public Node fieldsTransformationEmitCatalyst(DPLParser.FieldsTransformationContext ctx) {
         this.fieldsStep = new FieldsStep();
+        // if fields command has no arguments
+        if (ctx.getChildCount() == 1) {
+            throw new IllegalStateException(
+                    "fields command is missing field names, it requires at least one valid field name."
```
eemhu: would a check like this be possible instead?;

```java
if (ctx.fieldListType() != null && !ctx.fieldListType().fieldType().isEmpty()) {
            // ...
}
```

src/main/java/com/teragrep/pth_10/ast/FilteredColumns.java
```diff
@@ -0,0 +1,136 @@
...
+        else {
+            // if it is empty, the regexBuilder contains the final regex
+            regex = regexBuilder.toString();
+        }
+
+        final Pattern p = Pattern.compile(regex);
+        Matcher m;
+        final List<String> matchedFields = new ArrayList<>();
+
+        for (String column : columns) {
```
eemhu: `column` can also be final;

src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/FieldsTransformation.java
```diff
@@ -78,20 +78,28 @@ public Node visitFieldsTransformation(DPLParser.FieldsTransformationContext ctx)
...
-        if ("-".equals(oper)) {
-            StringListNode sln = (StringListNode) visit(ctx.fieldListType());
-            LOGGER.debug("Drop fields: stringListNode=<{}>", sln);
-
-            this.fieldsStep.setMode(AbstractFieldsStep.FieldMode.REMOVE_FIELDS);
-            this.fieldsStep.setListOfFields(sln.asList());
+        // if fields command has no arguments
+        if (
+            ctx.fieldListType() == null || ctx.fieldListType().fieldType() == null
+                    || ctx.fieldListType().fieldType().get(0).getText().isEmpty()
```
eemhu, line 85: Is it guaranteed that fieldType list is never empty?
`!ctx.fieldListType().fieldType().isEmpty()`

Abigael-JT: Not guaranteed, added a check for that

eemhu (approved):

PR #768, opened 2025-12-15
epoch migration command

### LINKED ISSUE #714: Implement a spark job to update missing logfile epoch metadataa
**Description**
Create a spark job to update missing archive meta data epoch columns using pth_06 to access the s3 files. Use the format and epoch from the result to update metadata for those logfiles without existing epoch values.
 
**Use case or motivation behind the feature request**
Goal is to have a robust spark job that can update the archive meta data and ensure all meta data  contains epoch values. 
 
**Related issues**
Required PTH_06 to support queries for a single file: https://github.com/teragrep/pth_06/issues/249
 
**Additional context**
Broad issue, ask for more information when required.

elliVM:
## Description

Implement support for a command to update missing epoch values of S3 object metadata in the archive SQL. Works by running the archive datasource in an epoch migration mode, where epoch value is fetched to the returned schemas `_time` column.

- Updates the pth_03 version to `9.4.0` with `teragrep exec migration epoch` command support
- Adds jooq library for dynamic SQL building
- Update the SQL logtime metadata of the S3 objects using epoch migration mode from the archiver
- Epoch migraiton mode of PTH_06 provides a calculated epoch value of the first even of the S3 in the _time column. In the _raw column there is a JSON format metadata about the event, this can be used to see how the event was parsed and what was the source of the _time column epoch.

### Testing
Included unit and dpl tests

### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality


src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/TeragrepEpochMigrationStep.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        EpochMigrationForeachPartitionFunction migrationFunction = new EpochMigrationForeachPartitionFunction(
+                config,
+                journalDBName
+        );
+        dataset.foreachPartition(migrationFunction);
+        return dataset;
+
+    }
+}
```
Tiihott: Add equals() and hashCode(). Tests should also be added.

Tiihott (changes requested):
Object equality tests should be added for some new classes.
Please rebase.

elliVM:
rebased

Tiihott (changes requested):
Tests fail when run via maven.

src/test/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationForeachPartitionFunctionTest.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public final class EpochMigrationForeachPartitionFunctionTest {
+
+    private final String user = "testuser";
```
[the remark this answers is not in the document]
elliVM: decided we should replace the singleton connection with a connection pool implementation, this will likely solve these issues with concurrency. Issue for connection pools #819

Tiihott (dismissed):
All tests pass and changes look fine, LGTM.

elliVM:
Doing testing in QA

Tiihott (dismissed):
Tests pass and new changes look ok. A new test run in QA should be done to check if the credential and JDBC driver fixes resolve the issues found in QA.

elliVM:
Successful migration in QA

Tiihott (dismissed):
Changes look fine and tests pass. LGTM

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    EpochMigrationBatchState accept(final Row row) {
+        final String rawString = row.getString(row.fieldIndex("_raw"));
+        final EventMetadata metadata = new EventMetadataFromString(rawString);
+        final long epoch = row.getTimestamp(row.fieldIndex("_time")).toInstant().getEpochSecond();
+        final String partitionString = row.getString(row.fieldIndex("partition"));
+        final long id = Long.parseLong(partitionString);
+
+        if (!metadata.isSyslog() && LOGGER.isInfoEnabled()) {
```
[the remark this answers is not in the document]
kortemik: It must be extracted properly from json in the _raw field. Also the object format should be updated in the same go

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
```diff
@@ -0,0 +1,144 @@
...
+                    .info(
+                            "Encountered non-syslog row id=<{}> using path extracted time value <{}> with precision of <{}> resulting epoch <{}>",
+                            id, metadata.pathExtracted(), metadata.pathExtractedPrecision(), epoch
+                    );
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Non-syslog row metadata: <{}>", metadata);
+            }
+        }
+        else if (LOGGER.isDebugEnabled()) {
+            LOGGER.trace("Updating partition id <{}> with epoch value <{}>", id, epoch);
```
kortemik, line 113: Trace is guarded by debug enabled, please change

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
```diff
@@ -0,0 +1,144 @@
...
+        final long id = Long.parseLong(partitionString);
+
+        if (!metadata.isSyslog() && LOGGER.isInfoEnabled()) {
+            LOGGER
+                    .info(
+                            "Encountered non-syslog row id=<{}> using path extracted time value <{}> with precision of <{}> resulting epoch <{}>",
+                            id, metadata.pathExtracted(), metadata.pathExtractedPrecision(), epoch
+                    );
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Non-syslog row metadata: <{}>", metadata);
```
kortemik: Payload must be logged only on the trace level

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationForeachPartitionFunction.java
```diff
@@ -0,0 +1,171 @@
...
+        this.connectionSource = connectionSource;
+        this.journalDBName = journalDBName;
+        this.batchSize = batchSize;
+        this.settings = settings;
+    }
+
+    @Override
+    public void call(final Iterator<Row> iter) {
+        try (final Connection conn = connectionSource.get()) {
+            if (conn.getAutoCommit()) {
```
kortemik: Why setting autocommit to false? Batch should commit only at the end and therefore autocommit tuning is extra?

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationForeachPartitionFunction.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    private BatchBindStep baseBatch(final DSLContext ctx) {
+        final Field<Long> epochField = DSL.field(DSL.name("epoch_hour"), Long.class);
+        final Field<Long> idField = DSL.field(DSL.name("id"), Long.class);
+        final Param<Long> epochParam = DSL.param("epoch_hour", Long.class);
+        final Param<Long> idParam = DSL.param("id", Long.class);
+        final Query baseQuery = ctx
+                .update(DSL.table(DSL.name(journalDBName, "logfile")))
```
kortemik, line 146: Update object_format as well

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EventMetadataFromString.java
```diff
@@ -0,0 +1,141 @@
...
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_10.steps.teragrep.migrate;
+
+import jakarta.json.JsonObject;
+
+import java.util.Objects;
+
+public final class EventMetadataFromString implements EventMetadata {
```
kortemik: This uses datacontroler logic meaning object is a data object. Proper way would be to create a list of possibilities and assing the json string from ctor and try if they are as parsed

kortemik (changes requested):

elliVM:
- added the object_format to the migration process.
- the object format names inSQL changed to match `OBJ_01`.
- logic to parse the JSON string in a correct way
- `ResolvedObjectFromats` to keep IDs for object formats available in memory, also adds new formats if missing

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EventMetadataFromString.java
elliVM: refactored this logic:
`EventMetadataFactory` tries to parse the string into known formats, if no matches or more than one an exception is thrown
`SyslogFormat` matches the JSON String to the expected result from pth_06 for a syslog format
`NonSyslogFormat` same for non syslog event from pth_06
`SyslogEvent` holds syslog event data
`NonSyslogEvent` holds non syslog event data

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationForeachPartitionFunction.java
elliVM: object format is not also migrated, format is added to the `object_format` table if missing. known formats have their IDs stored in a Map in `ResolvedObjectFormats` object so the ID can be directly accessed without lookup in database.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationForeachPartitionFunction.java
elliVM: removed setting autcommit to false

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
elliVM: fixed to trace level

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
elliVM: fixed guard to match the logger level

elliVM:
rebased

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EventMetadataFactory.java
```diff
@@ -0,0 +1,86 @@
...
+        this(Arrays.asList(new SyslogFormat(JSONString), new NonSyslogFormat(JSONString)));
+    }
+
+    EventMetadataFactory(final List<CandidateFormat> formats) {
+        this.formats = formats;
+    }
+
+    @Override
+    public EventMetadata get() {
+        final List<EventMetadata> validEvents = formats
```
kortemik: functional programming API is not preferred. please change to object oriented one.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EventMetadata.java
```diff
@@ -0,0 +1,70 @@
...
+ * Licensee must indemnify licensors and authors for any liability that these
+ * contractual assumptions impose on licensors and authors.
+ *
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_10.steps.teragrep.migrate;
+
+public interface EventMetadata {
```
kortemik: preferred name is ArchiveObjectMetadata

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EventMetadataFactory.java
```diff
@@ -0,0 +1,86 @@
...
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_10.steps.teragrep.migrate;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+final class EventMetadataFactory implements Supplier<EventMetadata> {
```
kortemik: I think this should be just a decorator to EventMetadata which takes in at ctor the list of possible formats and the EventMetadata. when changing to such design the decorator can directly implement the EventMetadata and return the format via format() by running the code which we have here in get()

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/CandidateFormat.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_10.steps.teragrep.migrate;
+
+import java.util.function.Supplier;
+
+public interface CandidateFormat extends Supplier<EventMetadata> {
```
kortemik: this interace is unnecessary, please return Stub / illegal argument exception from format() if not matching

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ParsedJson.java
```diff
@@ -0,0 +1,115 @@
...
+import jakarta.json.JsonException;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonReader;
+import jakarta.json.JsonStructure;
+import jakarta.json.JsonValue;
+
+import java.io.StringReader;
+import java.util.Objects;
+
+public final class ParsedJson {
```
kortemik: this is a json parsER object. instead it should be implementing EventMetadata directly

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedObjectFormats.java
```diff
@@ -0,0 +1,133 @@
...
+        this.resolvedObjectFormatsCache = resolvedObjectFormatsCache;
+    }
+
+    long resolve(final String objectFormat) {
+        return resolvedObjectFormatsCache.computeIfAbsent(objectFormat, this::resolveFromDatabase);
+    }
+
+    private long resolveFromDatabase(final String objectFormat) {
+        try {
+            // plain SQL since jooq can't insert into a generic table
```
kortemik: i am unsure what this comment is about? table is defined on L104 here?

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedObjectFormats.java
```diff
@@ -0,0 +1,133 @@
...
+import org.jooq.Record;
+import org.jooq.Table;
+import org.jooq.exception.DataAccessException;
+import org.jooq.impl.DSL;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+
+final class ResolvedObjectFormats {
```
kortemik, line 59: this is unnecessary and error-prone. One should do this as a transaction via inline or as a stored procedure so that database takes care of the proper id selection at the insert. 

```sql
START TRANSACTION;

INSERT INTO format (name)
VALUES ('rfc5424-syslog')
ON DUPLICATE KEY UPDATE id = LAST_INSERT_ID(id);

UPDATE foo (format_id)
VALUES (LAST_INSERT_ID()) WHERE id=123;

COMMIT;
```

this uses an upsert pattern.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/SyslogEvent.java
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_10.steps.teragrep.migrate;
+
+import jakarta.json.JsonObject;
+
+final class SyslogEvent implements EventMetadata {
```
kortemik: is this SyslogArchiveObject? Event is a naming mistake to avoid.

kortemik (changes requested):

elliVM:
changing to a temp table solution and using a stored procedure to update from temp table

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EventMetadata.java
elliVM: renamed

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EventMetadataFactory.java
elliVM: removed Supplier interface

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/CandidateFormat.java
elliVM: removed, stub object added where needed

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ParsedJson.java
elliVM: moved parsing logic into the `SyslogArchiveObjectMetadataFormat` and `UnknownArchiveObjectMetadataFormat` objects

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedObjectFormats.java
elliVM: object removed

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedObjectFormats.java
elliVM: transaction and object format id resolving was moved into a stored SQL procedure

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/SyslogEvent.java
elliVM: This object was removed, partial successor `SyslogArchiveObjectMetadataFormat` uses correct naming

elliVM:
rebased

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/SyslogArchiveObjectMetadataFormat.java
```diff
@@ -0,0 +1,114 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.StringReader;
+
+final class SyslogArchiveObjectMetadataFormat implements ArchiveObjectMetadataFormat {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SyslogArchiveObjectMetadataFormat.class);
+
+    @Override
+    public ResolvedFormat resolved(final String json) {
```
Tiihott: I believe it would be more OOP to first encapsulate the string via constructor and then call the resolved() method, as currently there is no encapsulation present in the class.
In other words format(json).resolved() instead of format().resolved(json).

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/UnknownArchiveObjectMetadataFormat.java
```diff
@@ -0,0 +1,112 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.StringReader;
+
+final class UnknownArchiveObjectMetadataFormat implements ArchiveObjectMetadataFormat {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UnknownArchiveObjectMetadataFormat.class);
+
+    @Override
+    public ResolvedFormat resolved(final String json) {
```
Tiihott: Same here, the String should be encapsulated first.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ArchiveObjectMetadataWithFormat.java
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+        if (validResolvedResults.size() != 1) {
+            throw new IllegalStateException(
+                    "Expected exactly one valid resolved result but got <" + validResolvedResults.size() + ">"
+            );
+        }
+        return validResolvedResults.get(0);
+    }
+}
```
Tiihott, line 100: equals() and hashCode() are missing.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedFormatImpl.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public String source() {
+        return source;
+    }
+
+    @Override
+    public boolean isStub() {
+        return false;
+    }
+}
```
Tiihott, line 171: equals() and hashCode() are missing.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/TeragrepEpochMigrationStep.java
```diff
@@ -0,0 +1,107 @@
...
+        });
+        final String streamingQueryName = "epoch-migration-stream-" + UUID.randomUUID();
+        final StreamingQuery streamingQuery = catCtx
+                .getInternalStreamingQueryListener()
+                .registerQuery(streamingQueryName, dataStreamWriter);
+        streamingQuery.awaitTermination();
+        return selectedColumnsDataset;
+
+    }
+}
```
Tiihott, line 131: equals() and hashCode() are missing.

Tiihott (changes requested):
There are some missing equals() and hashCode() overrides. Also the SyslogArchiveObjectMetadataFormat and UnknownArchiveObjectMetadataFormat classes should encapsulate something, for example the json String.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/TeragrepEpochMigrationStep.java
elliVM: added in [ad18c32](https://github.com/teragrep/pth_10/pull/768/commits/ad18c32bc7852fb21d84d8f4b162c72c5ae28b06)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedFormatImpl.java
elliVM: added [ad18c32](https://github.com/teragrep/pth_10/pull/768/commits/ad18c32bc7852fb21d84d8f4b162c72c5ae28b06)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ArchiveObjectMetadataWithFormat.java
elliVM: added [ad18c32](https://github.com/teragrep/pth_10/pull/768/commits/ad18c32bc7852fb21d84d8f4b162c72c5ae28b06)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/UnknownArchiveObjectMetadataFormat.java
elliVM: encapsulated in [ad18c32](https://github.com/teragrep/pth_10/pull/768/commits/ad18c32bc7852fb21d84d8f4b162c72c5ae28b06)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/SyslogArchiveObjectMetadataFormat.java
elliVM: encapsulated in [ad18c32](https://github.com/teragrep/pth_10/pull/768/commits/ad18c32bc7852fb21d84d8f4b162c72c5ae28b06)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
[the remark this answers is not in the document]
elliVM: object format to be also updated during the SQL stored procedure class

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedFormatImpl.java
```diff
@@ -0,0 +1,171 @@
...
+        else {
+            final ResolvedFormatImpl that = (ResolvedFormatImpl) o;
+            rv = Objects.equals(format, that.format)
+                    && Objects.equals(bucket, that.bucket) && Objects.equals(path, that.path) && Objects.equals(partition, that.partition) && Objects.equals(epoch, that.epoch) && Objects.equals(rfc5424Timestamp, that.rfc5424Timestamp) && Objects.equals(pathExtracted, that.pathExtracted) && Objects.equals(pathExtractedPrecision, that.pathExtractedPrecision) && Objects.equals(source, that.source);
+        }
+        return rv;
+    }
+
+    @Override
+    public int hashCode() {
```
Tiihott, line 164: EqualsVerifier test should be added

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedFormatStub.java
```diff
@@ -0,0 +1,118 @@
...
+            rv = false;
+        }
+        else {
+            rv = getClass() == o.getClass();
+        }
+        return rv;
+    }
+
+    @Override
+    public int hashCode() {
```
Tiihott, line 115: EqualsVerifier test should be added

Tiihott (changes requested):
Tests pass and changes look ok. But there are couple of EqualsVerifier tests missing from new changes.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedFormatStub.java
elliVM: added missing test in [0771084](https://github.com/teragrep/pth_10/pull/768/commits/077108480da209276fa650432f575a7538ee3ab2)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/ResolvedFormatImpl.java
elliVM: added missing test in [0771084](https://github.com/teragrep/pth_10/pull/768/commits/077108480da209276fa650432f575a7538ee3ab2)

Tiihott (dismissed):
LGTM

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/StubArchiveObjectMetadata.java
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_10.steps.teragrep.migrate;
+
+import java.util.Objects;
+
+public final class StubArchiveObjectMetadata implements ResolvedFormat {
```
kortemik: rename to ResolvedFormatStub ?

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/UnknownArchiveObjectMetadataFormat.java
```diff
@@ -0,0 +1,139 @@
...
+                        timestamp.getString("path-extracted-precision"),
+                        timestamp.getString("source")
+                );
+            }
+        }
+        catch (final RuntimeException e) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Error parsing JSON <{}> message <{}>", json, e.getMessage());
+            }
+            result = new StubArchiveObjectMetadata();
```
kortemik: assing stup as a field for this object `private static final stubArchiveObjectMetadata = new StubArchiveObjectMetadata();` and use `result = stubArchiveObjectMetadata`. this is an allocation optimization. stub is immutable and same instance can therefore used at all times. please document this to fkb

kortemik (changes requested):

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/UnknownArchiveObjectMetadataFormat.java
elliVM: Allocated only once in [4acfaee](https://github.com/teragrep/pth_10/pull/768/commits/4acfaee75977e50eff7ccf43bc6f25098b9da041)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/StubArchiveObjectMetadata.java
elliVM: renamed in [c0d0aeb](https://github.com/teragrep/pth_10/pull/768/commits/c0d0aeb31b6e8f1dd34fc15f6070dba2e8e587d2)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/UnknownArchiveObjectMetadataFormat.java
elliVM: Added documentation

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
```diff
@@ -0,0 +1,154 @@
...
+            final long batchCount,
+            final long acceptedRows
+    ) {
+        this.batch = batch;
+        this.batchSize = batchSize;
+        this.batchCount = batchCount;
+        this.acceptedRows = acceptedRows;
+    }
+
+    EpochMigrationBatchState accept(final Row row) {
```
kortemik: Would it be better approach to receive the Iterator<Row> iter in ctor and do the batch size management internally in store() call instead of using external control? At the moment EpochMigrationForeachPartitionFunction claims the responsibility of handling submission chunking.

kortemik (changes requested):

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
elliVM: I think that change would cause some tradeoffs. I designed the batch state object to be a pure immutable snapshot of a chunk of the data not the whole iterator. If we pass the iterator inside it the batch will be forced to handle the lifecycle, state and cursor of the whole data.

Issues I for see in this approach if I understood it correctly

1. Forces `EpochMigrationBatchState` to become mutable in some way,  since the iterator is statefull and iterating would require the state to change
2. Memory safety is harder to maintain if the iterator is consumed internally, if an iteration fails inside the state it would be harder to collect by the JVM GC. Also As the Spark's `ForeachPartitionFunction` works as an adapter here, if we keep the iterator in the same scope we can ensure that is a task fails spark can safely discard the entire memory context.
3. Harder to test since unit tests would require modeling of an full iterator instead of a single `Row`.

I would like to hear your input on these and if they are to be considered before implementing the change.

elliVM: please see this @kortemik

kortemik: The current approach uses a chain of EpochMigrationBatchState as a `Row decorator` to add database batch insert capability. It allocates already quite many new objects. It seems that there is Loader API with iterator defined in jOOQ/jOOQ/issues/4218 but I was unable to find good enough documentation for this so I guess this will do for now.

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationTempTable.java
```diff
@@ -0,0 +1,165 @@
...
+        this.epochHourField = epochHourField;
+        this.objectFormatField = objectFormatField;
+    }
+
+    /**
+     * Batch step for inserting values into epoch migration temp table
+     * 
+     * @return BatchBindStep for inserting logfile_id, epoch_hour and object_format into the epoch_migration_temp_table
+     */
+    BatchBindStep insertBatch() {
```
kortemik: please rename to batchBindStep();

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
```diff
@@ -0,0 +1,154 @@
...
+
+    BatchBindStep batch() {
+        return batch;
+    }
+
+    long totalAccepted() {
+        return acceptedRows;
+    }
+
+    EpochMigrationBatchState reset(final BatchBindStep newBatch) {
```
kortemik: this seems like a builder pattern. please justify why. otherwise this could be refactored so that one keeps the initial version stored outside the iteration loop and when batch is commited() then one could just start with the initial one. if the BatchBindStep however is mutable then this wouldn't work and if that is so the claim of immutability no longer holds and a better approach could be something else.

kortemik (changes requested):

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/StubResolvedFormat.java
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * To the extent this program is licensed as part of the Commercial versions of
+ * Teragrep, the applicable Commercial License may apply to this file if you as
+ * a licensee so wish it.
+ */
+package com.teragrep.pth_10.steps.teragrep.migrate;
+
+import java.util.Objects;
+
+public final class StubResolvedFormat implements ResolvedFormat {
```
kortemik: rename to ResolvedFormatStub

elliVM: renamed in [f1f1f39](https://github.com/teragrep/pth_10/pull/768/commits/f1f1f39af3cb384e47e759f70a0cf216c9747748)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationBatchState.java
elliVM: refactored to use the jooq loader api, this object is removed in [f1f1f39](https://github.com/teragrep/pth_10/pull/768/commits/f1f1f39af3cb384e47e759f70a0cf216c9747748)

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationTempTable.java
elliVM: removed method in [f1f1f39](https://github.com/teragrep/pth_10/pull/768/commits/f1f1f39af3cb384e47e759f70a0cf216c9747748)

elliVM:
Switched insertion into temp table to use jooq Loader api with a decorator to convert spark Rows into Object[] expected by the loader

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationIterator.java
```diff
@@ -0,0 +1,119 @@
...
+    public Object[] next() {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+        final Row row = origin.next();
+        final Timestamp ts = row.getTimestamp(row.fieldIndex("_time"));
+        if (ts == null) {
+            throw new RuntimeException("Column '_time' was null, cannot convert to epoch seconds");
+        }
+        final long epoch = ts.toInstant().getEpochSecond();
```
kortemik, line 87: this is not from resolved format?

kortemik, line 87: this is not rounded to epochHour?

kortemik (commented):
epoch is not from resolved format?

src/main/java/com/teragrep/pth_10/steps/teragrep/migrate/EpochMigrationIterator.java
eemhu: https://github.com/teragrep/pth_10/pull/768/commits/834d8b3905c527e2e4f6ae8b291efb9af644ad8b

eemhu:
tests pass locally

kortemik (approved):

PR #817, opened 2026-02-13
Improves error message for missing or invalid arguments for rangemap command

### LINKED ISSUE #689: Provide a better error message when user typoes rangemap command
**Description**
The current error message is too vague when user has a typo in query that has `rangemap` command.

````
java.lang.IllegalStateException: Parser failure on line 3, column 10 due to mismatched input '<EOF>' expecting COMMAND_RANGEMAP_MODE_FIELD
Please check that the query is written correctly. Otherwise, please report this error and include the query used and this error.
````

Instead, an error message that tells there's a typo or missing mandatory argument in the query would be more useful.

### PR DESCRIPTION
**Pull request by Abigael-JT**
Improves error message for missing or invalid arguments for rangemap command

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
+++ b/src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
@@ -80,6 +80,9 @@ public class RangemapTransformation extends DPLParserBaseVisitor<Node> {
         if (ctx.fieldType() != null) {
             this.rangemapStep.sourceField = new UnquotedText(new TextString(ctx.fieldType().getText())).read();
         }
+        else {
+            throw new IllegalArgumentException("rangemap command is missing required field parameter");
+        }
         return new NullNode();
     }
 
@@ -96,6 +99,11 @@ public class RangemapTransformation extends DPLParserBaseVisitor<Node> {
         // left side of range contains a trailing '-' character which needs to be removed
         if (valueLeft.endsWith("-")) {
             valueLeft = valueLeft.substring(0, valueLeft.length() - 1);
+        }
+        else {
+            throw new IllegalArgumentException(
+                    "Invalid range for rangemap command, the range should have both start and end values"
+            );
         }
 
         this.rangemapStep.attributeRangeMap.put(key, new String[] {
--- a/src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
+++ b/src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
@@ -89,6 +89,14 @@ public class RangemapTransformation extends DPLParserBaseVisitor<Node> {
     @Override
     public Node visitT_rangemap_attrnParameter(DPLParser.T_rangemap_attrnParameterContext ctx) {
         final String key = ctx.stringType().getText();
+
+        if (ctx.t_rangemap_rangeParameter().GET_RANGE_NUMBER_LEFT() == null) {
+            throw new IllegalArgumentException(
+                    "Invalid range values: <" + ctx.t_rangemap_rangeParameter().getText()
+                            + ">. rangemap command expects numerical range values"
+            );
+        }
+
         String valueLeft = ctx.t_rangemap_rangeParameter().GET_RANGE_NUMBER_LEFT().getText();
         final String valueRight = ctx
                 .t_rangemap_rangeParameter()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java [hunk 1]
```diff
@@ -80,12 +80,23 @@ public Node visitT_rangemap_fieldParameter(DPLParser.T_rangemap_fieldParameterCo
...
         }
+        else {
+            throw new IllegalArgumentException("rangemap command is missing required field parameter");
+        }
         return new NullNode();
     }
 
     @Override
     public Node visitT_rangemap_attrnParameter(DPLParser.T_rangemap_attrnParameterContext ctx) {
         final String key = ctx.stringType().getText();
```
elliVM, line 91: We are currently building separate tokens for `GET_NUMBER_LEFT` and `GET_NUMBER_RIGHT` which leads to partial parse states that force defensive handling here.

It could be worth considering refactoring the pth_03 parser grammar to represent 
`range :  NUMBER '-' NUMBER`. 

Then extraction becomes easy here
```
left = rangeCtx.NUMBER(0).getText()
right = rangeCtx.NUMBER(1).getText()

// parse to double
// ensure start <= end
```

elliVM (approved):
While we do fix some of the exception cases from parser in this change.
It seems to me that the visitor should not be responsible for handling parser syntax exceptions at all. It should interpret the parse tree with the assumption that the tree is structurally correct.

I think this exposes a common scenario in our current design where the pth_03 parser provides a broken or incomplete tree to the visitor. Result is that these parser failures leak into the visitors and further on into the exception messages.
Parser exception messages don't really make any sense to the end user as pointed out in the issue description. 

No easy solution for this here but I think it is worth raising the problem.

See my suggestion for refactoring the pth_03 rule and if it makes sense to you please create an issue for it.

src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
Abigael-JT: That's a valid point, I have raised a ticket for this in PTH_03 https://github.com/teragrep/pth_03/issues/120

src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
```diff
@@ -80,12 +80,23 @@ public Node visitT_rangemap_fieldParameter(DPLParser.T_rangemap_fieldParameterCo
...
+
+        if (ctx.t_rangemap_rangeParameter().GET_RANGE_NUMBER_LEFT() == null) {
```
eemhu: should the right number be checked as well?

Abigael-JT: yes that can be checked as well, added a check and a test

src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
```diff
@@ -104,8 +107,8 @@ public Node visitT_rangemap_attrnParameter(DPLParser.T_rangemap_attrnParameterCo
                 .GET_RANGE_NUMBER_RIGHT()
                 .getText();
 
-        // left side of range contains a trailing '-' character which needs to be removed
-        if (valueLeft.endsWith("-")) {
+        // left side of range contains a trailing '-' character which needs to be removed and right side must have a value
+        if (valueLeft.endsWith("-") && !valueRight.contains("missing")) {
```
eemhu, line 111: what is the missing check for?

src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
```diff
@@ -90,7 +90,10 @@ public Node visitT_rangemap_fieldParameter(DPLParser.T_rangemap_fieldParameterCo
     public Node visitT_rangemap_attrnParameter(DPLParser.T_rangemap_attrnParameterContext ctx) {
         final String key = ctx.stringType().getText();
 
-        if (ctx.t_rangemap_rangeParameter().GET_RANGE_NUMBER_LEFT() == null) {
+        if (
+            ctx.t_rangemap_rangeParameter().GET_RANGE_NUMBER_LEFT() == null
```
eemhu, line 94: I think t_rangemap_rangeParameter and t_rangemap_rangeParameter().rangeRightParameter() can be null

src/main/java/com/teragrep/pth_10/ast/commands/transformstatement/RangemapTransformation.java
Abigael-JT: Checks that the right value is not missing, if the right value is missing  valueRight is: "missing GET_RANGE_NUMBER_RIGHT".
Do you have a suggestion how this can be checked in a more readable  way?

PR #830, opened 2026-02-25
Improves error message when a field is converted twice

Abigael-JT:
## Description
Improves error message when user tries to convert the same field twice

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality

### DIFF
--- a/src/main/java/com/teragrep/pth_10/steps/convert/ConvertStep.java
+++ b/src/main/java/com/teragrep/pth_10/steps/convert/ConvertStep.java
@@ -53,7 +53,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


src/main/java/com/teragrep/pth_10/steps/convert/ConvertStep.java
```diff
@@ -90,7 +91,19 @@ public Dataset<Row> get(Dataset<Row> dataset) {
                     );
 
             // Get wildcarded fields
-            List<String> fields = getWildcardFields(cmd.getFieldParam(), rv.columns(), this.listOfFieldsToOmit);
+            final List<String> fields = getWildcardFields(cmd.getFieldParam(), rv.columns(), this.listOfFieldsToOmit);
+
+            fields.forEach(field -> {
+                if (fieldsToConvert.contains(field)) {
+                    throw new IllegalArgumentException(
+                            "Convert command allows converting a field with convert functions only once, the column < "
```
elliVM: there is extra whitespace after the `<` to be `< _time>`

eemhu (approved):

PR #886, opened 2026-08-05
convert toNumber UDF to return Double to support float values

### LINKED ISSUE #845: tonumber() can't convert floats
**Describe the bug**

If the value is a floating-point number, `tonumber()` returns nothing. 
 
**Expected behavior**

`tonumber()` should be able to convert `2.5` string value to a number. 
 
**How to reproduce**

```
| makeresults
| eval number="2.5"
| eval convert=tonumber(number)
```

If you try the following, it will return error instead of `2.5`:

`class java.lang.Double cannot be cast to class java.lang.String (java.lang.Double and java.lang.String are in module java.base of loader 'bootstrap')`

```
| makeresults
| eval convert=tonumber(2.5)
```
 
**Screenshots**

 
**Software version**

pth_10 version: 12.1.1
 
**Desktop (please complete the following information if relevant):**
 - OS: 
 - Browser: 
 - Version: 
 
**Additional context**

elliVM:
## Description

Resolves #845 

Makes the `tonumber(string)` DPL function support double values in addition to integers.

Spark columns have a fixed data type and cannot contain mixed numeric types (for example, `LongType` and `DoubleType` in the same column). Therefore, `DoubleType` is used as the return type to support both whole numbers and floating-point values.

Values such as `100` and `100.0` are numerically equivalent, so using `DoubleType` preserves the arithmetic behavior while allowing decimal values to be supported.

#### General

#### Assertions

#### Testing Data

#### Statements

#### Java

#### Other

### Code Quality


src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/ToNumber.java
```diff
@@ -56,34 +55,32 @@
...
+            }
+            else {
+                rv = (double) Long.parseLong(numberString, base);
+            }
         }
-        catch (NumberFormatException nfe) {
-            LOGGER.warn("Tonumber: Error parsing, returning 'null'. Details: <{}>", nfe.getMessage());
-            // Could not parse, return null
+        catch (final NumberFormatException exception) {
+            throw new RuntimeException(
```
elliVM: I don't think returning a null value was a good choice here, instead an exception should be thrown if the String number fails to parse.

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/ToNumber.java
```diff
@@ -56,34 +55,32 @@
...
 
-        // try parsing, otherwise return null
+        final double rv;
         try {
-            rv = Long.valueOf(numstr, base);
+            if (base == 10) {
+                rv = Double.parseDouble(numberString);
+            }
+            else {
+                rv = (double) Long.parseLong(numberString, base);
```
Laukkala: This means that we can only parse fractional numbers when they are in base 10.
I suppose it could be possible to write a custom parser that would support all the bases of toNumber uses, but is it out of scope for this issue?

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/ToNumber.java
```diff
@@ -56,34 +55,32 @@
...
                     "Tonumber: 'base' argument should be an integer value between 2 and 36."
             );
         }
 
-        // try parsing, otherwise return null
+        final double rv;
         try {
-            rv = Long.valueOf(numstr, base);
+            if (base == 10) {
+                rv = Double.parseDouble(numberString);
```
Laukkala: Can 'numberString' be null? If so, Double.parseDouble will throw a NullPointerException, unlike Long.ParseLong() which would throw a NumberFormatException in that case

src/test/java/com/teragrep/pth_10/evalTest.java
```diff
@@ -2261,17 +2261,50 @@ public void parseEvalTonumberNoBaseArgumentCatalystTest() {
...
+            Assertions.assertEquals(expectedSchema, res.schema());
+            final Dataset<Row> columnADataset = res.select("a").orderBy("a").distinct();
+            final List<Double> resultList = columnADataset
+                    .collectAsList()
+                    .stream()
+                    .map(r -> r.getDouble(0))
+                    .collect(Collectors.toList());
+            Assertions
+                    .assertEquals(Collections.singletonList(12.34), resultList, "result list should match the expected list");
         });
```
Laukkala, line 2309: Could add a test case for trying to use fractional values with bases other than 10
toNumber(1011.11,16) --> should result in an error indicating that fractional values using bases other than 10 are not supported.

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/ToNumber.java
Laukkala: Seems reasonable, null values could probably cause issues when chaining evals together.

Laukkala (commented):
Looks good overall, some comments related to what this method should support and null safety

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/ToNumber.java
elliVM: This is true. Correct radix conversion with fractional points outside of base 10 leads to IEEE 754 precision issues and non-standard syntax.  Use cases for this kind of operation are limited, so I would say it is outside of scope for a generic to number function.

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/ToNumber.java
elliVM: I think you are right, before the `nullValue` was returned in this case. I've added it back and made any `null` inputs return the `nullValue` early. This is against the single point of return but as base range check would also throw a NPE I think it is cleaner this way.

now then:
- null input -> output null, no NPE and works as spark UDFs are expected to work
- `String` that cant be parsed -> throw exception

[49120b7](https://github.com/teragrep/pth_10/pull/886/commits/49120b7e73d70e632cd190876ac939c27c5f7703)

src/test/java/com/teragrep/pth_10/evalTest.java
elliVM: added test in [49120b7](https://github.com/teragrep/pth_10/pull/886/commits/49120b7e73d70e632cd190876ac939c27c5f7703)

[ebdfc45](https://github.com/teragrep/pth_10/pull/886/commits/ebdfc45d592569358c1675c41b093b7dc3d6c723) adds mention of fractional values only accepting base 10

src/main/java/com/teragrep/pth_10/ast/commands/evalstatement/UDFs/ToNumber.java
Laukkala: Multiple returns could be resolved by adding the rest of the method inside an else block after checking the nulls, but I suppose its just a case of multiple returns vs if-else nesting

Laukkala (approved):
LGTM

eemhu:
@elliVM the other product does not return an error if parsing fails, rather does not return anything (from toNumber function). Could this be the case here as well?

elliVM:
Refactored to not return an error if parsing fails

[4db6c12](https://github.com/teragrep/pth_10/pull/886/commits/4db6c121ffd59cdb58e8a5aa0b6073baa5dcd9d4)

eemhu:
> Spark columns have a fixed data type and cannot contain mixed numeric types (for example, `LongType` and `DoubleType` in the same column). Therefore, `DoubleType` is used as the return type to support both whole numbers and floating-point values.

I think in MinMax UDF this was circumvented by using a StringType return. This of course then requires any proceeding command to parse the number

elliVM:
> > Spark columns have a fixed data type and cannot contain mixed numeric types (for example, `LongType` and `DoubleType` in the same column). Therefore, `DoubleType` is used as the return type to support both whole numbers and floating-point values.
> 
> I think in MinMax UDF this was circumvented by using a StringType return. This of course then requires any proceeding command to parse the number

I thought about this and reasoned that to have a `tonumber` UDF return a `StringType` could be confusing and would also move part of the parsing responsibility downstream. However we could do it this way if it is already done in the MinMax UDF.

elliVM:
coverted function to return `StringType`

elliVM:
rebased

